PR #3, opened 2021-10-13
Updates to parser to support compiler

Dashadower:
Parser now returns a format that's compatible with the compiler.

### DIFF
--- a/parser.py
+++ b/parser.py
@@ -1,13 +1,15 @@
 from typing import *
-from scanner import Token, Identifier, Operator, RealLiteral, IntLiteral, Special
+from scanner import Token, Identifier, Operator, RealLiteral, IntLiteral, Special, NullToken
 from ops import *
 
 # https://mc-stan.org/docs/2_18/reference-manual/bnf-grammars.html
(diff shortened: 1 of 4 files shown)


parser.py
```diff
@@ -159,39 +181,51 @@ def expressions(self):
...
+            self.remove()  # [
+            warnings.warn("Parser: Indices are assumed to be a single literal, not expression.")
             expressions = self.expressions()  # list of expression
+            self.expect_token(Special, "]")
             self.remove()  # ]
-            exp.index = Index(tuple(expressions))
+
+            # Assume index is a single identifier - this is NOT GOOD
+
+            exp.index = Index(tuple(expression.name for expression in  expressions))
```
Dashadower, line 224: Now looking at the compiler's code it seems like indexes atm can only be a single literal. Do you think it would be possible to support full expressions? It probably is overkill for a simple demo but if the goal is to support a full program with complete grammar then it should be considered.

bbbales2: I merged this PR cause it worked and it looks sick. We're going all the way from text to a probably very buggy model which is very exciting.

Two immediate issues are:
*. Figure out how lag operators should work
*. Figure out how parameter constraints should work

PR #11, opened 2021-10-24
Optimize

bbbales2:
Adding optimization option


Dashadower (approved):
Looks great. But maybe remove the model binaries.

PR #12, opened 2021-10-27
Constraints and some unary functions added to parser

Dashadower:
Looks like pytest works great! Parser can now handle `lower` and `upper` constraints specified after variable declaration. And not just fixed constants, but complete expressions.
Also added 5 unary functions to the parser/ops (exp, abs, floor, ceil, round) as well as the infix power operator(`^`).


bbbales2 (approved):
Very cool.

```
tau<lower= 0.0 + 5.0>
```

Math working in the expressions was unexpected lol

PR #13, opened 2021-11-02
Added parsing to a couple integration tests

bbbales2:


rat/model.py
```diff
@@ -25,7 +27,32 @@ class Model:
...
+    ):
+        data_names = data_df.columns
+        if model_string is not None:
+            if parsed_lines is not None:
+                raise Exception(
+                    "Only one of model_string and parsed_lines can be non-None"
+                )
+
+            parsed_lines = []
+            for line in model_string.splitlines():
```
Dashadower, line 44: One thing I might try is just wrap the entire multiline parsing into a function of the parser. This might be useful in the future in cases something changes and the spllit() works on different separators.

Dashadower (approved):
👍

PR #15, opened 2021-11-04
Multi index shift

bbbales2:


tests/integration/test_model_kalman.py
```diff
@@ -18,7 +19,7 @@ def test_optimize_kalman():
         ),
         ops.Normal(
             ops.Param("mu", ops.Index(("i",))),
-            ops.Param("mu", ops.Index(("i",), shift_columns=("i",), shift=1)),
+            ops.Param("mu", ops.Index(("i",), shift_columns=("i",), shifts=(1,))),
```
Dashadower, line 22: This looks fine, but how about unifying `shift_columns` and `shifts` into a just `shifts` argument? For example, if we have indexes `i, j, k` and want to shift `i` by 1 and `k` by 2, we can do something like: 
```
ops.Param("mu", ops.Index(("i", "j", "k"), shifts=(1, 0, 2)))
```

Dashadower (approved):

tests/integration/test_model_kalman.py
bbbales2: Hmm, there's a groupby that happens on the indices without the shift, and then you can use a zero to say don't do the groupby.

But I like the brevity. Maybe 'None' could indicate do the groupby on that column? So `shifts=(1, None, 2)` means groupby "j" and then shift the other two.

I can make this change tomorrow.

bbbales2: I switched to the `shifts = (1, 0, None)` style (so None means grouping applied).

I don't know what that means for the syntax, but the base example I have in my head is a random walk:

```
skill[team, year] ~ normal(skill[team, lag(year)], tau)
```

In this case you want to groupBy team before shifting (so you get that team's previous year's skill -- not some other teams). Having a syntax for not doing that grouping seems useful. I don't have an immediate suggestion on what would look nice (and also don't have a use yet)

Dashadower:
I'm merging this now so I can reflect it in the parser update branch.

PR #16, opened 2021-11-07
Update parser to support `shift()`

Dashadower:
parser now supports `shift()` that are called within brackets, with the respective signature: `shift(Identifier, IntLiteral)`
Identifier must be a data column.
Code is still a mess :(


Dashadower:
P.S. I think now Rat is officially not a context-free grammer anymore, since when parsing `[shift(...)]`, this requires knowledge that we are on the rhs of a statement, and `shift` only works given the index is within Data.

bbbales2:
> this requires knowledge that we are on the rhs of a statement, and shift only works given the index is within Data.

Oh lol, I see, we have a lot of context in there.

rat/variables.py
```diff
@@ -70,7 +70,7 @@ def compute_shifted_df(self, df, shifts):
 
         shifted_columns = []
         for column, shift in zip(shift_columns, shift_values):
-            shifted_column = grouped_df[column].shift(shift).reset_index(drop=True)
+            shifted_column = grouped_df[column].shift(int(shift)).reset_index(drop=True)
```
bbbales2, line 73: It looks like the shift values are being passed to the compiler as strings. If you remove this integer cast things fail (and it should work without this cast)

tests/integration/test_model_kalman.py
```diff
@@ -47,32 +52,34 @@ def test_optimize_kalman():
...
-            ops.RealConstant(0.0),
-            ops.RealConstant(1.0),
-        ),
-    ]
-
-    model = Model(data_df, parsed_lines)
+    model_string = """
+    score_diff ~ normal(skills[team1, year] - skills[team2, year], sigma);
+    skills[team, year] ~ normal(skills[team, shift(year, 1)], 0.5);
+    sigma<lower = 0.0> ~ normal(0, 1.0);
```
bbbales2, line 58: It works!

bbbales2 (changes requested):
Found one problem adding tests

rat/variables.py
Dashadower: Changed so parser returns int

tests/integration/test_model_kalman.py
Dashadower: 🎉🎉🎉

rat/ops.py
```diff
@@ -43,7 +43,7 @@ def __str__(self):
 @dataclass
 class Index(Expr):
     names: Tuple[str]
-    shifts: Tuple[Union[str, None]] = None
+    shifts: Tuple[Union[str, None]] = (None,)
```
Dashadower, line 46: I've changes this line because `all(shift is None for shift in shifts)` didn't work on  `variables.Index`.

Dashadower: @bbbales2

bbbales2 (approved):

PR #18, opened 2021-11-15
Non center integration test

bbbales2:


tests/integration/test_model_noncenter.py
```diff
@@ -0,0 +1,67 @@
...
+
+    model_string = """
+    y ~ normal(theta[school], sigma[school]);
+    theta[school] = mu + z[school] * tau;
+    z[school] ~ normal(0, 1);
+    mu ~ normal(0, 5);
+    tau<lower = 0.0> ~ lognormal(0, 1);
+    """
+
+    ## TODO: Add a unit test that the thing above parses to the thing below
```
Dashadower, line 23: If you need  to test parser, you can create the tree yourself, and then check if their string representations match like [so](https://github.com/bbbales2/regressions/blob/4e73d419e36e9e4d548e29fa3bb716f36d1d0c07/tests/unit/test_parser.py#L103)
```
assert statement.__str__() == expected.__str__()
```

Dashadower (approved):
LGTM

PR #23, opened 2021-11-21
Docs

bbbales2:
This pull adds pdoc as automatic documentation generation and renders them with github pages. I did [pdoc](https://pdoc.dev/docs/pdoc.html) cause it looks clean and simple and there's a github action to automatically update stuff.

The docs are generated automatically on merge to master. They aren't generated on each push to the branch, so the html diffs don't show up in the PR cause I thought that might just get annoying. If it turns out that it'd be more convenient to have the html files to check, then we can switch to that.

I also changed the autoformatting stuff to only run on merge to master. Our PRs won't be automatically re-written all the time which might be convenient? I'm not sure -- anyway, the main branch will stay autoformatted and we don't have to worry about that so much.

You can see how the docs get rendered here: https://bbbales2.github.io/regressions/


Dashadower (approved):
Sweet! More motivation to write docstrings!

PR #28, opened 2021-11-22
Sampler ess rhat

bbbales2:
Looks like ESS and Rhat are easy to compute with Arviz

I need to add tests for these.

Still working on getting a reasonable use of blackjax in place.


Dashadower (approved):
LGTM. On a side note, does blackjax automatically thin samples?

bbbales2:
> automatically thin samples

Nah, we'd have to do that ourselves but wouldn't be difficult.

PR #29, opened 2021-11-22
Non center parameterizations and assignments

Dashadower:
(This is still work in progress, will mention when it's ready)

This branch pretty much tries to make assignments work, which enables hierarchical non-centered parameterization in the form of:

```
y[n] ~ normal(theta[n], sigma);
theta[n] = mu + tau * unit_normal[n];

unit_normal[n] = normal(0, 1);
tau<lower=0> ~ normal(0, 2);
mu ~ normal(0, 5);
sigma = 1;
```


Dashadower:
Oh darn didn't see #6 lol, moving there.

PR #30, opened 2021-11-23
Update parser and ops.Index to support union() in subscripts

Dashadower:
Now `Index.name` is a tuple of tuples or strings, with a nested tuple denoting union of its content.

```
tau<lower = 0.0>[union(home_team, away_team), year]
```
is the way to use it. However, 
```
tau[union(home_team, away_team), year]<lower = 0.0>
```
will not work.


Dashadower:
@bbbales2 Please check that you're fine with this syntax. I'm going to merge this into `non_center` so I can tweak the compiler.

PR #32, opened 2021-11-29
Add docs/make improvements for scanner and parser

Dashadower:
Because someone who writes compilers are probably going to freak out if they read my code :)
Goals:
- Implement the placeholder math functions and add some new ones
- Verbose error messages for Parser/Compiler
- Rewrite portions of the Parser that's not optimal.
- Write extensive docs + package introduction writeup


Dashadower:
It turns out while improving parser I happen to add some more math functions to rat.

Dashadower:
```
E       rat.parser.CompilationError: An error occured during processing the following line:
E       score_diff ~ normals(skills[home_team, year]-skills[away_team, year],sigma);
E                    ^
E       
E       Expected token types ['Identifier'] with value in ['normal', 'bernoulli_logit', 'log_normal', 'cauchy'], but received Identifier with value 'normals'!
```

Now we have some cool looking errors!

bbbales2:
Yoooo that error message looks awesome

rat/__init__.py
```diff
@@ -1,3 +1,145 @@
...
+    - `rat.ops.Ceil` : `ceil(x)`
+    - `rat.ops.Round` : `round(x)`
+    - `rat.ops.Sin` : `sin(x)`
+    - `rat.ops.Cos` : `cos(x)`
+    - `rat.ops.Tan` : `tan(x)`
+    - `rat.ops.Arcsin` : `arcsin(x)`
+    - `rat.ops.Arccos` : `arccos(x)`
+    - `rat.ops.Arctan` : `arctan(x)`
+    - `rat.ops.Logit` : `logit(x)`
+    - `rat.ops.InverseLogit` : `inverse_logit(x)`
```
bbbales2, line 140: hype hype hype

rat/__init__.py
```diff
@@ -1,3 +1,145 @@
...
+5         6         122         117       BOS       BKN         5.0  2016
+6         7         109          91       TOR       DET        18.0  2016
+7         8          96         107       MIL       CHA       -11.0  2016
+8         9         102          98       MEM       MIN         4.0  2016
+9        10         102         107       NOP       DEN        -5.0  2016
+10       11          97         103       PHI       CLE        -6.0  2016
+```
+and we're trying to estimate some hidden skill parameter for each team. We can write the following model:
+```
+score_diff ~ normal(skills[home_team, year] - skills[away_team, year], sigma);
```
bbbales2, line 27: I've been thinking we should require subscripts on data as well. So here instead of `score_diff` we have `score_diff[home_team, away_team, year]`. Just for an abundance of clarity.

Dashadower: Hmm. Is it just for defining which subscripts we're going to reference on the rhs?

bbbales2: Yeah something like that was the idea. Thinking more the problem is that there'd be a lot of wrong ways to index this. And I don't even think we have the syntax to make it clear what is going on. Imagine if we had multiple measurements for a single home_team, away_team, year tuple (which I guess wouldn't happen in basketball but maybe something else)

New stance is if the variable is in the dataframe, there are no subscripts.

bbbales2:
Not sure why the tests weren't running/docs weren't building. I marked as ready for review and merged main in.

rat/__init__.py
Dashadower: >  Imagine if we had multiple measurements for a single home_team, away_team, year tuple

I don't understand how this can be a problem. Wouldn't we just increment the log probability of said variable multiple times?

bbbales2: Everywhere we use subscripts so far they're refer to unique elements of data structure.

By allowing subscripts on data, using the above data, we could say something like:

```
y[year] ~ ...
```

y is a column of the dataframe, so the value of y on each row is just the value of y on each row. The index is not adding anything. It makes it seem like there are only num(years) elements of y which probably isn't true.

We currently use these subscripts on the left hand side for parameters. In this case they're defining names which expressions on the right hand side can use. Maybe it makes sense to allow no subscripts on those as well in the case that the subscripts aren't used on the right hand side.

Like:
```
score_diff ~ normal(skills[year], sigma);
skills ~ normal(0, 1);
```

We might already allow this in the code -- I don't think I've thought through it tho lol.

bbbales2 (approved):
Is this ready to merge? Looks good to me

Dashadower:
> Is this ready to merge? Looks good to me

I still have some WIP stuff that I haven't finished, but we can merge as-is and I can start another PR.

PR #37, opened 2021-12-13
Prototype RNG interface

bbbales2:
I put together some stuff for rngs. The goal here is to do prior/posterior predictives.


bbbales2:
You know what, I think we can probably just get by without rngs.

The thing is it should be really easy to do RNGs outside of rat. With rat, you can just generated the parameters you'll need on the outside with transformed parameters.

This also conveniently let's us sidestep the awkwardness that is number of random draws from rng doesn't need to match draws from MCMC.

Dashadower:
> You know what, I think we can probably just get by without rngs.
> 
> The thing is it should be really easy to do RNGs outside of rat. With rat, you can just generated the parameters you'll need on the outside with transformed parameters.
> 
> This also conveniently let's us sidestep the awkwardness that is number of random draws from rng doesn't need to match draws from MCMC.

Yeah that makes sense, doing simulations outside without the user having to specify the structure with rng functions sound better.

Issue #39, opened 2021-12-21
Multiple line programs don't parse correctly


bbbales2:
Might just be a problem with how we're passing strings to the parser, see [here](https://github.com/bbbales2/regressions/blob/main/rat/model.py#L132)

bbbales2:
Here's a program that'll cause the current setup to go into infinite loop:
```
y ~ bernoulli_logit(
  a +
  b
);
```

Dashadower:
Yeah right now it uses just newline as separation characters. I'm working on some scanner modifications which better implements the grammar.

Dashadower:
I'm closing this since I'm certain this has been fixed. Feel free to reopen if it happens again.

PR #40, opened 2021-12-22
Scanner and Parser improvements

Dashadower:
Features:

- Scanner rewrite - a complete state machine with well-defined transitions and states. Super easy to maintain and update in the future.
- Multiline model strings(#39). Now scanner receives an arbitrary number of lines, but will return a list equal to the "true" model line count.
- Precedence-based parser. Thanks to Pratt Parsing, operator precedence is now added for arithmetic and prefix ops. Also a new exception for the Parser, which allows verbose error messages like the Scanner.
- Exponential Distribution added(beta parameterization, like stan)
- A bit more unit tests for parser and scanner.


rat/compiler.py
```diff
@@ -330,14 +330,14 @@ def compile(data_df: pandas.DataFrame, parsed_lines: List[ops.Expr]):
...
-                    assert isinstance(upper, ops.RealConstant)
+                    #assert isinstance(lower, ops.RealConstant)
+                    #assert isinstance(upper, ops.RealConstant)
 
                     if lhs_key in variable_indexes:
                         subexpr = variables.Param(lhs_key, variable_indexes[lhs_key])
                     else:
                         subexpr = variables.Param(lhs_key)
-                    subexpr.set_constraints(lower.value, upper.value)
+                    subexpr.set_constraints(float(eval(lower.code())), float(eval(upper.code())))
```
Dashadower, line 340: @bbbales2 
I've enabled the constraints to be expressions instead of just `RealConstant`. Not a great way to make this work; suggestions welcome.

rat/compiler.py
```diff
@@ -331,14 +331,14 @@ def compile(data_df: pandas.DataFrame, parsed_lines: List[ops.Expr]):
                 elif isinstance(lhs, ops.Param):
                     lower = lhs.lower
                     upper = lhs.upper
-                    assert isinstance(lower, ops.RealConstant)
-                    assert isinstance(upper, ops.RealConstant)
+                    # assert isinstance(lower, ops.RealConstant)
+                    # assert isinstance(upper, ops.RealConstant)
```
bbbales2, line 335: Delete comments

rat/compiler.py
bbbales2: What is this making work? This is like saying it's okay to have expressions as constraints as long as we can evaluate them?

grammar.md
```diff
@@ -107,7 +110,7 @@ constraint_range ::= 'lower' '=' expressioni ',' 'upper' = expression
                    | 'lower' '=' expression
                    | 'upper' '=' expression
 
-Distribution ::= ("normal")
+Distribution ::= ("normal", "bernoulli_logit", "log_normal", "cauchy")
```
bbbales2, line 113: Also exponential with this PR!

tests/integration/test_model_noncenter.py
```diff
@@ -23,7 +23,19 @@ def eight_schools_model():
...
+    parsed_lines = [
+        ops.Normal(ops.Data("y"), ops.Param("theta", ops.Index(("school",))), ops.Data("sigma")),
+        ops.Assignment(
+            ops.Param("theta", ops.Index(("school",))),
+            ops.Sum(ops.Param("mu"), ops.Mul(ops.Param("z", ops.Index(("school",))), ops.Param("tau"))),
+        ),
+        ops.Normal(ops.Param("mu"), ops.RealConstant(0.0), ops.RealConstant(5.0)),
+        ops.Normal(ops.Param("z", ops.Index(("school",))), ops.RealConstant(0.0), ops.RealConstant(1.0)),
+        ops.LogNormal(ops.Param("tau", lower=ops.RealConstant(0.0)), ops.RealConstant(0.0), ops.RealConstant(1.0)),
+    ]
```
bbbales2, line 36: Can delete these ops now. This model is compiling from the string version of the model above.

tests/integration/test_rat_cli.py
```diff
@@ -10,7 +10,7 @@
 from rat import compiler
 from rat.model import Model
 from rat.parser import Parser
-from rat.scanner import scanner
+from rat.scanner import Scanner
```
bbbales2, line 13: Looks like unused import

tests/integration/test_sample.py
```diff
@@ -9,7 +9,7 @@
 from rat import compiler
 from rat.model import Model
 from rat.parser import Parser
-from rat.scanner import scanner
+from rat.scanner import Scanner
```
bbbales2, line 12: Looks like unused import

tests/unit/test_parser.py
```diff
@@ -101,6 +100,13 @@ def test_parser_rhs_index_shift_multiple():
     assert statement.__str__() == expected.__str__()
 
 
+def test_parser_invalid_statement():
+    with pytest.raises(Exception):
```
bbbales2, line 104: Should match part of the error string of this exception (to make sure the expected exception is actually being thrown).

An example from https://docs.pytest.org/en/6.2.x/assert.html is:

```python
def test_match():
    with pytest.raises(ValueError, match=r".* 123 .*"):
        myfunc()
```

tests/unit/test_parser.py
```diff
@@ -20,43 +23,39 @@ def test_parser_multiple():
...
-        line = line.lstrip().rstrip()
-        if not line:
-            continue
-        print(f"----running for line [{line}]")
-        print(line)
-        print("{", [(x.__class__.__name__, x.value) for x in scanner(line)], "}")
-        print(Parser(scanner(line), data_names, line).statement())
+    scanned_lines = Scanner(input_str).scan()
+    for line in scanned_lines:
+        print(Parser(line, data_names, line).statement())
```
bbbales2, line 28: I don't think there's anything really wrong with print statements in here, but pytest eats them by default so they won't be terribly useful unless you're running pytest with prints turned on.

tests/unit/test_scanner.py
```diff
@@ -0,0 +1,145 @@
+import pytest
+from rat.scanner import *
+
+
+def test_scanner_invalid_scientific():
+    input_str = "g<lower=1.4ef> ~ normal(0.0, 1.0);"
+    parsed = Scanner(input_str)
+    with pytest.raises(TokenizeError):
```
bbbales2, line 8: Since this is a custom Exception I don't think we need to add a match, but it still might be useful.

bbbales2 (changes requested):
Looks good! Couple things before merge

rat/compiler.py
Dashadower: Yup, they should be known at compile time. Although right now there are no checks for that.

Dashadower:
I was outside all-day today so I couldn't work on the changes XD. Thanks for doing it instead.

Dashadower:
It looks like I can't reply or answer your question on your review; I left my comment on my original comment thread. Feel free to resolve and merge.

bbbales2 (approved):

PR #49, opened 2021-12-30
Updating docs

bbbales2:


bin/rat
```diff
@@ -20,6 +20,7 @@ parser.add_argument("--num_warmup", type = int, default = 1000, help = "Number o
 parser.add_argument("--num_draws", type = int, default = 1000, help = "Number of draws to sample after warmup (only available for sample)")
 parser.add_argument("--retries", type = int, default = 4, help = "Number of times to retry optimization (only available for optimize)")
 parser.add_argument("--tolerance", type = float, default = 1e-2, help = "All optimization results must be within rtol = atol = tolerance of each other or throw an error (only available for optimize)")
+parser.add_argument("--target_acceptance_rate", type = float, default = 0.85, help = "Target acceptance rate for adaptation")
```
bbbales2, line 23: Along with updating the documentation I changed an argument to sample to allow setting the target accept rate

rat/__init__.py
```diff
@@ -59,16 +196,156 @@
...
+`output` we can do:
+
+```
+rat mrp.rat mrp.csv output
+```
+
+Type `rat -h` for full usage info.
+"""
+
+__all__ = ["model", "fit"]
```
bbbales2, line 346: I only import these two modules by default so these are the only two pdoc renders docs for

rat/model.py
```diff
@@ -33,7 +33,7 @@ class Model:
     parameter_offsets: List[int]
     parameter_sizes: List[Union[None, int]]
 
-    def constrain(self, unconstrained_parameter_vector, pad=True):
+    def _constrain(self, unconstrained_parameter_vector, pad=True):
```
bbbales2, line 36: I added an underscore to a bunch of these internal functions so they won't show up as external.

PR #51, opened 2021-12-30
Support comments in model code

Dashadower:
Scanner takes care of removing comments from the model code. It simply will ignore trailing characters after a `#` and won't generate tokens until newline


bbbales2 (approved):
Looks good!

PR #52, opened 2021-12-30
Doc3

bbbales2:


rat/__init__.py
```diff
@@ -101,11 +101,21 @@
 
 ## Likelihood
 
-Rat inherits from the regression languages. As such, the first line of a Rat
-program is the likelihood. This line is special because it can access named
-variables in the input dataframe.
+Every Rat program comes with a dataframe. The dataframe defines the data which
+the Rat model fits itself too.
```
Dashadower, line 105: I really hate being a pedant but I think this should be 'to'.

Dashadower (approved):
Looks good!

rat/__init__.py
bbbales2: Nah nah, you're right. I think pedantic implies a bit over the top, this is just a good ol' fashion grammar fix.

PR #41, opened 2021-12-23
Example models

bbbales2:

Issue #44, opened 2021-12-27
Use of unassigned variable on right hand side


bbbales2:
This model doesn't work (I was trying to define a non-centered parameterization and accidentally wrote it):

```python
df = pandas.DataFrame({ "abortion" : [0], "state" : [1] })

model_string = """
abortion ~ bernoulli_logit(a_state[state]);
a_state[state] = a_state[state];
"""

model = Model(df, model_string=model_string)
```

Error looks like:
```
 return {self.line.rhs.code()}"])
  File "/home/bbales2/regressions/rat/ops.py", line 104, in code
    variable_code = self.variable.code()
AttributeError: 'NoneType' object has no attribute 'code'
```

Dashadower:
Compiler assumes all variables on RHS are known at evaluation time. We would need a default prior like for #43 and in addition require changes to the dependency graph generation, since we now need ordering between lines that share the same variable on the LHS. e.x. `a_param[group] ~ normal(0, 1.0);` should be evaluated before `a_param[group] = a_param[group] * 2;`

bbbales2:
I agree this should just be an error -- no need to support the syntax here -- just meant to say the error was confusing in figuring out what was happening.

bbbales2:
This is fixed. Error is now:

```
rat.exceptions.CompileError: An error occurred while compiling the following line(2:18):
a_state[state]' = a_state[state];
                  ^
Exactly one (no more, no less) subscripts can be shifted in a recursively assigned variable
```

It's a correct error though I probably could have worded it a bit clearer lol.

Issue #43, opened 2021-12-26
No priors on parameters throw not-super-informative errors


bbbales2:
Here's code. Remove the two commented lines to see the two different errors:

```python
df = pandas.DataFrame({ "abortion" : [0], "state" : [1] })

model_string = """
abortion ~ bernoulli_logit(a_state[state]);
a_state[state] = a_state_z[state] * sigma_state;
a_state_z[state] ~ normal(0, 1.0); # Remove this line and you'll see an error about the index for a_state_z not existing
sigma_state<lower = 0.0> ~ normal(0.0, 1.0); # Remove this line and you'll still get an error but a different one
"""

model = Model(df, model_string=model_string)
```

I think parameters with no priors should throw warnings. We could go pedantic and require a prior, but seems simpler to just log a warning.

Here is one of the errors as an example:
```
  File "/home/bbales2/regressions/rat/compiler.py", line 271, in compile
    parameter_base_df[target_op_name].columns = tuple(target_op_index_key)
TypeError: 'NoneType' object is not iterable
```

Dashadower:
Hmm. I think removing 
```
a_state_z[state] ~ normal(0, 1.0);
```
should throw an error. The assignment line does assume a_state_z exists. IMO not requiring a prior equates to implicit declaration which looks more prone to coding errors. But I guess if warnings suffice then I'm on board.

With that said when implementing, I think this would become "if a variable on RHS is not present, immediately create a parameter with said name and subscript that follows some default distribution".

bbbales2:
I don't think we require priors on parameters yet. Anyway this can close and we can come back whenever.

PR #57, opened 2022-01-07
Compiler Refactor

Dashadower:
Sorry for another ginormous commit. This commit has PR #56 merged in, so merging this will also merge said PR.

Features:
- refactored `rat.compiler.compiler()` into `rat.compiler.Compiler()` class that breaks down the previously big function into smaller ones. You can now access the compile results through `Compiler(data_df, expr_tree, model_code).compile())`
- Topological ordering of expression tress as to just variables. Previously we had to run a nested for loop with the outer along target variables and inner along code lines/trees. Now the trees themselves are ordered, which allows just a single loop and no possibility of N! worst case running time.
- Better compile error messages and detection. Compiler also has detailed error messages much like the Parser and Scanner. In addition, a semi-full rewrite resulted in less dangling edge cases, so a standard python exception being raised when there's something wrong with the code wouldn't happen frequently.
```
rat.compiler.CompileError: An error occured while compiling the following line(2:21):
theta[school] = mu + z[schools] * tau;
                     ^
Couldn't index RHS variable z's declared subscript('schools',) from LHS base df, which has subscripts ('school',).
```
```
rat.compiler.CompileError: An error occured while compiling the following line(1:49):
score_diff ~normal (skills[ home_team , year ] - skillss[away_team, year],sigma[year]);
                                                 ^
Could not find variable skillss
```
```
rat.compiler.CompileError: An error occured while compiling the following line(2:0):
skills[team, year] ~ normal(skills_mu[year], 1);
^
Could not topologically order the expression tree. This is because your model is cyclic.
```


Dashadower:
This PR fixes #41, #43, #44

bbbales2:
Yo this looks sick.

How does this error work though:
```
rat.compiler.CompileError: An error occured while compiling the following line(1:49):
score_diff ~normal (skills[ home_team , year ] - skillss[away_team, year],sigma[year]);
                                                 ^
Could not find variable skillss
```

In my mind not finding a variable isn't an error -- the compiler should just infer one exists cause the use.

rat/ops.py
```diff
@@ -21,6 +24,8 @@ def __str__(self):
 @dataclass(frozen=True)
 class RealConstant(Expr):
     value: float
+    line_index: int = -1  # line index of the original model code
+    column_index: int = -1  # column index of the original model code
```
bbbales2, line 28: I think we can put the defaults in the base class (and make Expr a dataclass). There's a trick to make this work, but I think it should be possible: https://stackoverflow.com/a/58525728.

Dashadower:
> Yo this looks sick.
> 
> How does this error work though:
> 
> ```
> rat.compiler.CompileError: An error occured while compiling the following line(1:49):
> score_diff ~normal (skills[ home_team , year ] - skillss[away_team, year],sigma[year]);
>                                                  ^
> Could not find variable skillss
> ```
> 
> In my mind not finding a variable isn't an error -- the compiler should just infer one exists cause the use.

Well, if you mean it shouldn't raise an error because not defining a prior(LHS use) for a variable doesn't mean it doesn't exist, then yes it should be changed to add a default prior like Uniform(-2, 2). I think this is what you meant.

But currently the assumption is that in order for a variable to exist it must be used at least once on the LHS. Since the lines are topologically ordered if we haven't seen `skillss` that means it wasn't used on LHS and thus is a non-existant variable.

Dashadower:
What should be the default prior be? Uniform(-2, 2) like stan?

bbbales2:
> Well, if you mean it shouldn't raise an error because not defining a prior(LHS use) for a variable doesn't mean it doesn't exist

Oh oh, I see, so like a prior hasn't been defined for this. I can live with priors being required for now. It's not obvious to me that we don't want something like this eventually.

In this case could the error be something like `No prior found for skillss`?

Dashadower:
> > Well, if you mean it shouldn't raise an error because not defining a prior(LHS use) for a variable doesn't mean it doesn't exist
> 
> Oh oh, I see, so like a prior hasn't been defined for this. I can live with priors being required for now. It's not obvious to me that we don't want something like this eventually.
> 
> In this case could the error be something like `No prior found for skillss`?

Yeah I think that's a better error message.

But it wouldn't be too hard to throw in a default prior in the current code. [This line](https://github.com/bbbales2/regressions/blob/49253f6b0dd6a113f8faeff79fe1e3dd51569f4c/rat/compiler.py#L506) is the only thing that we have to change; just create a `variable.Param` and a simple linefunction with the default prior.

bbbales2:
I don't think default priors for now. For now priors required is good, and in the future we might want to consider improper priors (just don't do anything), and I haven't really thought so much about actual default priors in the way that rstanarm has them. Even if we had default from data priors, I might want to make them like:

```
mu ~ rstanarm_default();
```

To at least have an indicator that something is happening.

bbbales2 (approved):
This all looks very nice. Good cleanup. I'm gonna try to use the default values for the Expr class. If that works, I'll add it and merge this all. If it doesn't, I'll just merge this as-is.

rat/ops.py
Dashadower: Sounds good. I was quite frustrated having to copy-paste those two lines repeatedly.

bbbales2: Cool beans. It's some Python 3.10 stuff that makes this easy: https://stackoverflow.com/a/69822584

PR #58, opened 2022-01-08
Starting to work on compiling a model to a Python module

bbbales2:


bbbales2:
Here's what the 8-schools code compiled to a Python file looks like:

```python
import rat.constraints
import rat.math
import jax

unconstrained_parameter_size = 10

def constrain_parameters(unconstrained_parameter_vector, pad=True):
    unconstrained_parameters = {}
    parameters = {}
    total = 0.0

    # mu                                                                                                                                           
    unconstrained_parameters['mu'] = unconstrained_parameter_vector[..., 0]
    parameters['mu'] = unconstrained_parameters['mu']

    # tau                                                                                                                                          
    unconstrained_parameters['tau'] = unconstrained_parameter_vector[..., 1]
    parameters['tau'], constraints_jacobian_adjustment = rat.constraints.lower(unconstrained_parameters['tau'], 0.0)
    total += jax.numpy.sum(constraints_jacobian_adjustment)

    # z                                                                                                                                            
    unconstrained_parameters['z'] = unconstrained_parameter_vector[..., 2 : 10]
    parameters['z'] = unconstrained_parameters['z']

    return total, parameters

def transform_parameters(data, subscripts, parameters):
    # theta                                                                                                                                        
    parameters['theta'] = parameters['mu'] + parameters['z'][subscripts['school__2']] * parameters['tau']

    return parameters

def evaluate_densities(data, subscripts, parameters):
    target = 0.0
    target += jax.numpy.sum(jax.scipy.stats.norm.logpdf(parameters['mu'], 0, 5))
    target += jax.numpy.sum(rat.math.log_normal(parameters['tau'], 0, 1))
    target += jax.numpy.sum(jax.scipy.stats.norm.logpdf(parameters['z'][subscripts['school__0']], 0, 1))
    target += jax.numpy.sum(jax.scipy.stats.norm.logpdf(data['y'], parameters['theta'][subscripts['school__3']], data['sigma']))

    return target
```

rat/variables.py
```diff
@@ -190,14 +190,15 @@ def size(self):
...
-        return f"assigned_param__{self.ops_param.name}"
+        return f"{self.ops_param.name}"
 
 
 @dataclass
 class SubscriptUse:
     names: Tuple[str]
     df: pandas.DataFrame
     subscript: Subscript
+    unique_id: str
```
Dashadower, line 201: Is there a reason why this is necessary?

Dashadower, line 201: What is this attribute used for?

bbbales2: In a program you can have different subscript uses with the same shifts and names that correspond to differerent dataframes.

So like:
```
y ~ normal(mu[group], sigma[group]);
mu[group] ~ normal(somethingelse[group], ...);
...
```

I'm now saving subscript uses in a global data structure, so I need to make sure they have their own name for generated code.

I thought about putting this unique_id generation as part of the SubscriptUse class so we didn't need to pass a unique_id in, but decided to use leave it in the compiler code itself.

Dashadower: Oh okay that makes sense 👍

Dashadower (approved):

PR #60, opened 2022-01-15
Adding nuts sampler

bbbales2:


rat/nuts.py
```diff
@@ -0,0 +1,381 @@
...
+                    metrics[0] += number_waiting_threads
+                    metrics[1] = 0.1 * (time.time() - start) / number_waiting_threads + 0.9 * metrics[1]
+
+                    self.waiting_threads[active_thread] = False
+
+                    break
+
+            # There's a race condition here that hopefully doesn't lead to errors. If a non-running thread doesn't
+            # get to wait before the running thread sets and clears the event, it will wait for the next round to run
+            with self.context_lock:
```
Dashadower, line 133: For every step, record the values where you want to calculate the gradient values for. And then a single thread calculates the gradients for each chain. Once it's finished it returns the values back to each chain thread. Is this correct or did I understand it wrong?

bbbales2: Yeah -- it's a synchronization mechanism where every thread that has signaled it is active (by doing the activate_thread context manager) lines up to have gradients computed, but only one thread makes the actual call.

But I'm about to change this and try something else.

bbbales2: I removed the thread synchronization. Now the Potential class is more or less pre-allocating memory to make copies fast.

Dashadower (approved):
Looks Great!
In `nuts.py`, is it okay to use standard numpy arrays with jax? Also the RNG stuff, IIRC jax needs its own version tossed in.

bbbales2:
> is it okay to use standard numpy arrays with jax

I think for the most part jax knows how to convert these automatically. We do it when we call gradients, for instance. Is there a place in particular that seems weird?

> Also the RNG stuff, IIRC jax needs its own version tossed in.

All the random number stuff is handled in numpy now, though it leaks into the Rust. All the jax stuff should be deterministic (given the numpy randomness).

Dashadower:
> > is it okay to use standard numpy arrays with jax
> 
> I think for the most part jax knows how to convert these automatically. We do it when we call gradients, for instance. Is there a place in particular that seems weird?
> 
> > Also the RNG stuff, IIRC jax needs its own version tossed in.
> 
> All the random number stuff is handled in numpy now, though it leaks into the Rust. All the jax stuff should be deterministic (given the numpy randomness).

Ah I see, thanks for the clarification.

PR #61, opened 2022-01-18
WIP typing system and constant folding

Dashadower:
and also recursion


Dashadower:
## Prototype typing system:
- Integer (element of type Numeric)
- Real (element of type Numeric)
- Subscript

### when it screws up:
Data columns' "type" differ depending on where they are used. For example, using a data column named x like the following:
```
y ~ normal(x * 5, 1.0);
```
suggests we're actually using the values of x, and hence x has type `Numeric`.
However, say we use data columns as a subscript:
```
y ~ normal(mean_param[group], 1.0);
```

Now `group` denotes the set of elements in column group, which should have type `SubscriptType`, which denotes the index set.

rat/ops.py
```diff
@@ -60,14 +107,87 @@ def get_key(self):
...
+        Combines multiple subscripts into a single subscript
+        """
+        names: List[str] = []
+        shifts: List[Union[Expr, None]] = []
+        for subscript in self.subscripts:
+            subscript = subscript.fold()
+            shifts.extend(subscript.shifts)
+            names.extend(subscript.names)
+
+        return Subscript(names=tuple(names), shifts=tuple(shifts), line_index=self.line_index, column_index=self.column_index)
```
bbbales2, line 144: Oh interesting so what is happening here is for an expression like:

```
variable[A, B, C]
```

This is parsed as:
```
variable[subscript_A, subscript_B, subscript_C]
```

And then we fold things together to be:
```
variable[subscript_A_B_C]
```

This about right?

bbbales2: Actually I see the folding happens through basically everything -- is the folding done exclusively to bundle the subscripts together? Or are there other types of constants?

Dashadower: > Oh interesting so what is happening here is for an expression like:
> 
> ```
> variable[A, B, C]
> ```
> 
> This is parsed as:
> 
> ```
> variable[subscript_A, subscript_B, subscript_C]
> ```
> 
> And then we fold things together to be:
> 
> ```
> variable[subscript_A_B_C]
> ```
> 
> This about right?

Yes. Same goes for shifts. We bundle multiple subscripts into a single one.

Dashadower: > Actually I see the folding happens through basically everything -- is the folding done exclusively to bundle the subscripts together? Or are there other types of constants?

Yup for constants too. If you check the parser unittests you can see some examples.

rat/ops.py
```diff
@@ -108,6 +240,35 @@ def scalar(self):
...
+            self.subscript = self.subscript.fold()
+        if not isinstance(self.lower, (RealConstant, IntegerConstant)):
+            raise ConstantFoldError(
+                f"Lower bound value must fold into a Numeric constant at compile time, but folded expression {self.lower} is not a constant!",
+                self.line_index,
+                self.column_index,
+            )
+        if not isinstance(self.upper, (RealConstant, IntegerConstant)):
+            raise ConstantFoldError(
+                f"Upper bound value must fold into a Numeric constant at compile time, but folded expression {self.upper} is not a constant!",
```
bbbales2, line 256: Oh I see so the constraint checks move now that the parser knows types!

bbbales2: Well I guess this isn't the constraint itself, but the checking that it is indeed a constant thingy

Dashadower: Yup. we can just call `fold()` and check.

Dashadower: The drawback for this is perhaps when the compiler needs to throw an error - the column index can be a bit messed up

Dashadower:
@bbbales2 I'm doing the recursion work on this branch. Will that be fine? (SV model is merged in)

bbbales2:
Sounds good I'll close the PR with just the example since it's here now.

Dashadower:
Closed for #65

PR #64, opened 2022-02-12
Types + Prime syntax + Recursive assignment + Constant folding at compile time

bbbales2:


tests/integration/test_model_errors.py
```diff
@@ -43,6 +44,32 @@ def test_optimize_missing_subscript():
...
+    with pytest.raises(Exception, match="The primed uses must come last"):
+        model = Model(data_df, model_string=model_string)
+
+
+def test_lines_in_wrong_order_for_assignments():
+    data_df = pandas.read_csv(os.path.join(test_dir, "bernoulli.csv"))
+
+    model_string = """
+    y' ~ bernoulli_logit(mu[group]);
+    mu[group]' = 0.1 + mu2[group]; # If you remove the '0.1 +' this fails to parse
```
bbbales2, line 64: The compiler gives a type error if you remove the `0.1 +` part of this line. Not sure what is going on, but presumably it can't infer the type for mu2?

Dashadower: I left out a type signature for assignment. Added in the new commit.

rat/ops.py
```diff
@@ -24,77 +47,196 @@ def __str__(self):
...
     variable: variables.SubscriptUse = None
 
     def get_key(self):
         return self.names
 
-    def code(self):
-        return f"subscripts['{self.variable.code()}']"
+    def code(self, scalar = False):
+        if scalar:
+            return f"subscripts['{self.variable.code()}'][index]"
```
bbbales2, line 109: All the ops now take an extra argument `scalar` which indicates if they should do code generation assuming they're being evaluated like scalars or not.

If they're scalars, then assume the variable `index` is the scalar index.

tests/integration/test_kalman.py
```diff
@@ -1,33 +1,53 @@
...
+    assert mu_df["mu"].to_list() == pytest.approx(mu_ref, rel=1e-2)
+
+
+def test_optimize_time_series_non_center():
+    data_df = pandas.read_csv(os.path.join(test_dir, "kalman.csv"))
+
+    model_string = """
+    y' ~ normal(mu[i], 0.1);
+    mu[i]' = mu[shift(i, 1)] + epsilon[i];
+    epsilon[i] ~ normal(0, 0.3);
```
bbbales2, line 49: This is the new thing we can do.

rat/compiler.py
```diff
@@ -523,13 +587,59 @@ def _generate_python_source(self, ordered_expr_trees: List[ops.Expr]):
...
+
+                                number_subscripts_shifted = sum(shift is not None for shift in shifts)
+                                if number_subscripts_shifted != 1 or shifts[0] is None:
+                                    raise Exception(
+                                        "Internal compiler error (shouldn't alert here): When shifting an assigned variable, can only have one shifted subscript and it must be the first"
+                                    )
+
+                                carry_size = max(carry_size, shifts[0])
+
+                    if lhs_used_in_rhs:
```
bbbales2, line 625: When we're using the variable assigned on the left hand side on the right hand side, we know we need to evaluate everything with a scan.

For the code:
```
y' ~ normal(mu[i], 0.1);
mu[i]' = mu[shift(i, 1)] + epsilon[i];
epsilon[i] ~ normal(0, 0.3);
```

We'll get a transformed parameters block like:
```python
ef transform_parameters(data, subscripts, parameters):
    # mu
    def scan_function_6(carry, index):
        next_value = carry[index - subscripts['i__1__3'][index] - 1] + parameters['epsilon'][subscripts['i__4'][index]]
        next_value_as_array = jax.numpy.array([next_value])
        return jax.numpy.concatenate([next_value_as_array, carry[1:]]), next_value

    _, parameters['mu'] = jax.lax.scan(scan_function_6, jax.numpy.zeros(1), jax.numpy.arange(10))

    return parameters
```

rat/compiler.py
```diff
@@ -523,13 +587,59 @@ def _generate_python_source(self, ordered_expr_trees: List[ops.Expr]):
...
+
+                        if carry_size == 0:
+                            raise Exception(
+                                "Internal compiler error (shouldn't alert here): Carry size should be greater than zero when scan used"
+                            )
+
+                        scan_function_name = f"scan_function_{self._get_unique_identifier()}"
+
+                        code += f"    def {scan_function_name}(carry, index):\n"
+                        code += f"        next_value = {top_expr.rhs.code(scalar = True)}\n"
```
bbbales2, line 636: When we're doing a scan, make sure and ask for scalar code generation.

rat/ops.py
```diff
@@ -108,15 +250,111 @@ def scalar(self):
...
+
+    def code(self, scalar=False):
         variable_code = self.variable.code()
         if self.subscript:
-            return f"parameters['{variable_code}'][{self.subscript.code()}]"
+            if self.assigned_by_scan:
+                if scalar == False:
+                    raise Exception("Internal Error: Cannot do a vector evaluation of an assigned variable")
+
+                return f"carry[index - {self.subscript.code(scalar)} - 1]"
```
bbbales2, line 296: Assume that when this variable is marked that it will be assigned by a scan that there will be a variable `carry` available that will be saving the carry values. This assumes things are sorted in a certain way and whatnot.

bbbales2 (commented):
I'm thinking now that recursive stuff and match are actually separate things.

As long as we carry along the idea that out-of-bounds accesses are zeros, we can do recursive things without match statements (the termination condition for the recursion comes from where the variable is used other than the recursion, so that sticking point is deflected).

Now I'm thinking that since we can skip match for now (and still do some non-centered time series parameterizations) we do.

There are a couple sizable questions even if we leave match out:
1. How bad is the performance hit from this?
2. How to handly groupbys? Previously shift did groupbys on every subscript that wasn't shifted -- I'm not sure that works now.

For groupbys there is:
1. How to write them
2. The relationship between the primary dataframes and the secondary dataframes and how this changes with groupbys
3. How to execute scans when there are groupbys

For 1, what I like is if we change our bracket operator so something like:
```
[a, b, c]
```

means subscript over `a`, `b`, `c` in the regular way, and then:
```
[a; b, c]
```

means subscript over `a` for each group indicated by `b` and `c`. A groupby and shift would be:
```
[shift(a, 1), b; c]
```

For performance reasons (how the carry is computed in assigned statements), I think we should only allow one shifted subscript for now.

I think groupby operations should be allowed at least on the primary dataframe. It might make sense to allow separate groupbys on secondary dataframes.

How to compile these to python source is a question too -- it would be very nice to write everything so it still compiles quickly. The simple solution of doing a for loop over groups would lead to slow compile times I believe.

rat/ops.py
```diff
@@ -108,15 +250,111 @@ def scalar(self):
...
 
     def __str__(self):
-        return f"Param({self.name}, {self.subscript.__str__()}, lower={self.lower}, upper={self.upper})"
+        return f"Param({self.name}, subscript={self.subscript.__str__()}, lower={self.lower}, upper={self.upper}, assigned={self.assigned_by_scan})"
+
+
+@dataclass
+class Match(Expr):
+    first: Expr
+    otherwise: Expr
```
bbbales2, line 309: When I moved match to be an expression I changed it to be more like an if-else just to make things simple on myself. If we pull out match, I'll just move all the stuff to a different PR so we have it in the future.

bbbales2:
> How bad is the performance hit from this?

Still need to test this

> How to handly groupbys? Previously shift did groupbys on every subscript that wasn't shifted -- I'm not sure that works now.

For now have decided to try to just match the current standard. That means no new syntax, and group by all the columns that are not shifted.

bbbales2:
For the record, looks like the performance hit of using a scan is around 10x. It looks like scan performance is pretty sensitive to how much unrolling there is too (unrolling 2 times was faster than 1 or 3 for me)

rat/variables.py
```diff
@@ -151,10 +153,24 @@ class Param:
     subscript: Subscript = None
     lower: float = float("-inf")
     upper: float = float("inf")
+    constraints_set: bool = False
 
     def set_constraints(self, lower, upper):
```
Dashadower, line 158: how does `set_constraints` work with `Data`?

rat/variables.py
```diff
@@ -107,36 +105,40 @@ def get_numpy_indices(self, df):
             "__index"
         ].to_numpy(dtype=int)
 
+    def get_first_in_group_indicators(self, shifts):
```
Dashadower, line 108: Can you explain this code bit?

rat/compiler.py
```diff
@@ -523,13 +580,58 @@ def _generate_python_source(self, ordered_expr_trees: List[ops.Expr]):
...
+
+                                carry_size = max(carry_size, shifts[-1])
+
+                                # There will be groupings if there is more than one shift
+                                if len(shifts) > 1:
+                                    need_first_in_group_indicator = True
+
+                    if lhs_used_in_rhs:
+                        # We need to scan in this case
+                        scan_function_name = f"scan_function_{self._get_unique_identifier()}"
```
Dashadower, line 619: lol that's alot of cryptic looking fstrings

Dashadower: I wonder if we can make the code generation code tidier. With the indents and stuff it looks hard to read and keep track.

Dashadower (approved):
Very neat scan implementation <3

rat/variables.py
bbbales2: I wasn't thinking about constraints on data. We could certainly add those. For now I'm assuming no mistakes :P.

rat/compiler.py
bbbales2: Lol yeah at the f-strings. I should probably add some example comments. Much harder to read than what they actually are.

I agree this code feels kinda at its limits. I'm not quite sure how to do the code-gen more cleanly though, so my thought is to just wait until this code fails to do anything more complex (since I'm not sure in what direction to take it). If you can think of a way to improve, go for it.

rat/variables.py
bbbales2: Yeah, for any recursive assignment where there is more than one subscript, the non-shifted subscripts will be grouping variables.

When scanning, each group should be computed independently of each other. To achieve this we reset the carry when we index into a new group (other option would be a for loop over all groups, or a vmap across all groups -- the first would be slow to compile and I'm not quite sure how to implement the second though I'm interested). Look at the `jax.lax.cond` statement in the following code. This works because the groups are sorted (if all the different groups were jumbled together, the reset wouldn't work):

```python
def transform_parameters(data, subscripts, first_in_group_indicators, parameters):
    # skills
    def scan_function_7(carry, index):
        carry = jax.lax.cond(first_in_group_indicators['skills'][index], lambda : (0.0), lambda : carry)
        (carry0) = carry
        next_value = carry0 + parameters['epsilon'][subscripts['team_year__4'][index]]
        return (next_value), next_value

    _, parameters['skills'] = jax.lax.scan(scan_function_7, (0.0), jax.numpy.arange(15))

    return parameters
```

rat/variables.py
Dashadower: Oh sorry github diff viewer made it look like `set_constraints` existed for `Data`. I agree with you

PR #65, opened 2022-02-19
Types and folding into `prime`

Dashadower:
- Implements typing and elementary constant folding. 
- Removes some duplicate code from the parser.
- Adds Stochastic volatility model from `sv`
- Match statement is parsed into `MatchedStatement`


rat/scanner.py
```diff
@@ -214,6 +228,14 @@ def raise_error(self, msg):
...
 
     def delimeter_state(self):
+        bracket_dict = {")": "(", "}": "{", "]": "["}
+        if self.current_char in bracket_dict.values():
+            self.bracket_stack.append(self.current_char)
+        elif self.current_char in bracket_dict.keys():
+            if len(self.bracket_stack) == 0:
+                self.raise_error(f"Found unmatching brackets {self.current_char}")
+            elif self.bracket_stack.pop() != bracket_dict[self.current_char]:
+                self.raise_error(f"Found unmatching brackets {self.current_char}")
```
bbbales2, line 238: Is the bracket_stack for better errors or is it also doing something else and the errors are convenient?

Either way exciting to have errors like this!

rat/parser.py
```diff
@@ -310,76 +316,144 @@ def expect_token(
...
-            expressions.append(expression)
-            shift_amounts.append(shift_amount)
+                expression = self.expression(is_subscript=is_subscript)
+                expressions.append(expression)
+
+        return expressions
+
+    def parse_nud(self, is_lhs=False, is_subscript=False):
+        token = self.peek()
+        if isinstance(token, RealLiteral):  # if just a real number, return it
```
bbbales2, line 355: Oh man this is a top tier replace with match statements candidate!

rat/parser.py
```diff
@@ -585,26 +611,99 @@ def expression(self, min_precedence=0, is_lhs=False):
...
             op = self.peek()
-            if AssignmentOps.check(op):
+            if op.value == "=" and self.peek(k=1).value == "match":
+                # Matched assignment block
+                reverse_order = False
+                self.remove()  # assignment operator
+                self.remove()  # "match"
+                self.expect_token(Identifier)
+                bounded_variable = self.peek()
+                if bounded_variable.value == "rev":
```
bbbales2, line 637: Oh wow you already included this nice

rat/parser.py
```diff
@@ -585,26 +611,99 @@ def expression(self, min_precedence=0, is_lhs=False):
...
+                        "Match statement missing else(recurrence equation) stataement!", self.model_string, lhs.line_index, lhs.column_index
+                    )
+                return_statement = MatchedStatement(
+                    lhs=lhs,
+                    initial_declarations=initial_declarations,
+                    recurrence_equation=recurrence_equation,
+                    bounded_variable_name=bounded_variable.value,
+                    reverse_order=reverse_order,
+                    line_index=lhs.line_index,
+                    column_index=lhs.column_index,
```
bbbales2, line 695: Excellent excellent I believe this is all the info necessary

rat/ops.py
```diff
@@ -647,6 +1076,19 @@ def __iter__(self):
...
+    def __post_init__(self):
+        signatures = {
+            (types.RealType,): types.RealType,
+        }
+        self.out_type = types.get_output_type(signatures, (self.subexpr.out_type,))
+
+    def fold(self):
+        self.subexpr = self.subexpr.fold()
+        if isinstance(self.subexpr, RealConstant):
+            return RealConstant(value=jax.scipy.special.logit(self.subexpr.value))
```
bbbales2, line 1088: I wonder if these should be jax operations or since they execute in the compiler they should be numpy. If tests pass with this, let's roll with this but this might turn into a problem in the future (not clear to me when using jax stuff what is being copied to device memory or what)

bbbales2 (approved):
Wooo

rat/scanner.py
Dashadower: It's for error checking. The bracket checking stuff was in the parser, but it turns out the scanner does a single pass of the entire model string so the bracket checking is free if done on the parser.

rat/parser.py
Dashadower: The sv rat file has the syntax specifications I based on.

rat/parser.py
Dashadower: I'm not sure what keywords other than `first` would be needed for the initial declarations.

rat/ops.py
Dashadower: Since the folding is just for scalars, it should be fine to use numpy.

PR #67, opened 2022-03-15
[WIP]Remove folding

Dashadower:
# DO NOT MERGE

MRP with 10k iterations shows:
323.95it/s without folding
273.30it/s with folding

So I'm rolling back the folding changes, but keeping the typechecks.

### DIFF
--- a/rat/compiler.py
+++ b/rat/compiler.py
@@ -498,22 +498,22 @@ class Compiler:
                         if assigned_key == symbol_key:
                             symbol.assigned_by_scan = True
 
-                            if symbol.subscript is None or all(shift is None for shift in symbol.subscript.shifts):
+                            if symbol.subscript is None or all(shift == 0 for shift in symbol.subscript.shifts):
                                 msg = f"Recursively assigning {symbol_key} requires a shifted subscript on the right hand side reference"
                                 raise CompileError(msg, self.model_code_string, symbol.line_index, symbol.column_index)
 
                             shifts = symbol.subscript.shifts
 
-                            if sum(shift is not None for shift in shifts) != 1:
+                            if sum(shift != 0 for shift in shifts) != 1:
                                 msg = "Exactly one (no more, no less) subscripts can be shifted in a recursively assigned variable"
                                 raise CompileError(msg, self.model_code_string, symbol.line_index, symbol.column_index)
 
-                            if shifts[-1] is None:
+                            if shifts[-1] == 0:
                                 msg = "Only the right-most subscript of a recursively assigned variable can shifted"
                                 raise CompileError(msg, self.model_code_string, symbol.line_index, symbol.column_index)
 
-                            if any(shift <= 0 for shift in shifts if shift is not None):
-                                msg = "All shifts in a recursively assigned variable must be greater than zero (equal to zero is not supported)"
+                            if any(shift < 0 for shift in shifts):
+                                msg = "All shifts in a recursively assigned variable must be greater than or equal to zero"
                                 raise CompileError(msg, self.model_code_string, symbol.line_index, symbol.column_index)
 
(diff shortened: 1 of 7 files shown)


Dashadower:
Okay, there are some interesting observations I found while messing around with types and shifts.

First, like I said on discord I decided to remove the folding stuff since XLA does it already, and does it better than I do.

Now, I took at look at `shift`. Right now, the default value for `ops.Subscript.shifts` when there's no shift in place is `None`. Which does make sense, because it literally means there's no shifting done there. But I think having a shift value of `0` instead of `None` makes more sense meaning-wise. I think it helps to view it this way when doing type checks for subscript: it should have a signature like so: `(names: index_set, shifts: integer) -> index_set`. Subscripts map values in the data dataframe(say `CLE`, `IND`, ... for baseball teams) to indexes in the parameter dataframe. So checking if `shifts` is an integer, instead of `None` or Integer gives more clarity.

---

Which gives way to another problem: If I were to define the role of the AST, it would be to purely represent the syntax without any "interpretation" of values(e.g. calling `int(eval())` on a constant). Currently, shift values are stored as python integers, but I think at least in the AST tree, it should be represented as a `IntegerConstant` object. 

Current implementation:
```
Param
└── Subscript
    ├── names (type: python str)
    └── shifts (type: python int)
```
should be changed to:
```
Param
└── Subscript
    ├── names (ast.Subscript with type: rat.types.SubscriptSetType)
    └── shifts (ast.IntegerConstant with type: rat.types.IntegerType)
```
`ast.Subscript` doesn't exist, but it's there to differentiate how the type of a data column changes depending on its usage: `RealType` if used as `Data`, `SubscriptSetType` if used as a subscript.

---

If you think about codegen, the actual AST nodes(in `rat.ops`) that get transformed are: `Data`, `Param`, `Subscript`, `IntegerConstant`, and `RealConstant` (I'm gonna call them primitive five). The first 3 get transformed into dataframe indexing operations, and the latter two get `eval()` applied. I think it makes more sense to create IR nodes for these 5 nodes and merge `rat.variables` equivalents' code into them, since right now, actual code gen for these values rely on both `code()` methods in ops and `variables.py`. This may be helpful since we only have a single class dictating how code is generated.

So to summarize:
- rename current `ops.py` to `ast.py`
- Strip the code gen in `ops` for primitive five and merge with `rat.variables`' into new IR nodes.
- AST of primitive 5 only exist as a shell that contain the literals.


The parse/compilation step would also change:
1. Parse into naive AST (`parser.py`)
2. Do type checking with AST (`parser.py` + `ast.py`)
3. Replace AST nodes of primitive five with their IR equivalents (`compiler.build_base_df, compiler.build_variables`)
4. Run code gen

All in all, I think if designed well, this should drastically simplify the python codegen stuff.

bbbales2:
This makes sense. My worries are:

1. I don't see how it makes the Python code generation [here](https://github.com/bbbales2/regressions/blob/main/rat/compiler.py#L518) much simpler than it is already. The model problem, I think, would be how does it make the scan code generation simpler? Either that or constraints. Both of those generate extra lines of code that are not sorta 1-to-1 coming from the AST.

2. The variables are doing the work of local variables -- there'd need to be a replacement structure to hold these things. This is how I think of variables differently from the ops.

I think right now the code works by, generate the ops as an AST, parse through the AST a bunch building variables and IR, but write the IR data back into the original ops AST instead of building a separate structure.

bbbales2:
The idea of separating the expressions and values makes a lot of sense.

I guess the other worry I'll add is that now every operation appears in three separate places:

1. It must be parsed
2. It must be AST'd
3. It must be compiled

Ops is already kinda the longest file.

Dashadower:
Closing for #69

PR #68, opened 2022-03-19
Io rework

bbbales2:
This adds a thin argument and also outputs info on leapfrog steps taken and whether or not the U-turn happened because of a divergence.

For thinning, I make it so that num_draws still determines the output size, just internally we'll sample num_draws * thin things to get that output.


Dashadower (approved):
LGTM

PR #69, opened 2022-03-20
AST + Codegen Refactor

Dashadower:
## AST and code generation refactor
---
### Design Goals
- Separate code generation logic with model representation
- Convert compilation into a generic process, so that multiple code generation backends can utilize generic functions provided by the baseline compiler.

### Key Changes
1. `ops.py` -> `ast.py`, `codegen_backends.py`.
  Separate code generation from the AST representation into a separate module. Instead of each AST node containing its codegen implementation, they are moved into a codegen class that implements its own logic, which is called by a visitor pattern.
2. `compiler.py` - symbol table
  Instead of iterating through the AST tree in whichever fashion the codegen needed, the compiler now defaults to generating a symbol table that stores information about each variable. The symbol table is now used to drive a a generic subscript generator function that a codegen backend can invoke by passing its primary and target variable information. The hope is that it will be easier to build purpose-specific codegen backends without having to collect/manage base df information.

### Places for improvement/unsatisfactory points
- in `codegen_backends.py`, code is generated by invoking visitor functions, which append the node's code to a class attribute string. This is hard to identify at a single glance, since for example, code which was previously: 
  ```model_string += f"{lhs.code()} + {rhs,code()}"```
  would now be represented as:
```
lhs.accept(self)
self.expression_string += " + "
rhs.accept(self)
```
- Symbol tables do separate generating subscript arrays from codegen, but there may be a need for additional AST tree passes during codegen, which isn't easy to access from the codegen backend class.
- Stateful codegen backends. Codegen backends need states, since the codegen behavior of a node can differ by ex) whether the param is at LHS or RHS. Currently, the easiest way is to use class attributes within the backend class or passing variables along the visitor function call chain. They are less intuitive to read as to just using a loop-based traversal codegen.
- Using the visitor pattern + generic functions mean a lot more boilerplate code and redundant code iterations. For example, previously we can check for subscript errors and build base dfs within the same loop. Breaking the base df generation into a function of its own means needing another code iteration to perform checks.


Dashadower:
Important point that I'm not sure 100% about:
When converting from an unconstrained vector into a constrained parameter, if pad is needed(shift is present), I defaulted to padding a 0 to the end of the parameter array which I think is the previous compiler's behavior. So for a length 4 variable that needs padding:
```
unconstrained     = [1, 2, 3, 4]
parameters["a"]   = [1, 2, 3, 4, 0]  # 0 padded at the end
```
then the indices for shift would be `[4, 1, 2, 3]`, and when accessing the parameter, it would be **passed with the pad excluded** like so: `parameters["a"][:-1]`.

I'm not sure if this would be the correct way to do it. But passing the parameter with the pad included will throw an `IndexError` for unmatching shapes.

rat/compiler.py
```diff
@@ -7,82 +7,333 @@
...
+    by the programmer.
+    unconstrained_vector_start_index: For parameters, the start index of the unconstrained parameter vector. Generated
+    by `SymbolTable.build_base_dataframes()`.
+    unconstrained_vector_end_index: For parameters, the end index of the unconstrained parameter vector. Generated by
+    `SymbolTable.build_base_dataframes()`.
+    pad_needed: true if a padding is needed for the variable
+    """
+
+    name: str
+    variable_type: VariableType
```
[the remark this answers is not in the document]
bbbales2: Maybe it's that we don't necessarily know the type at construction?

rat/compiler.py
```diff
@@ -498,207 +577,155 @@ def _prepare_for_scans(self):
...
-        Dict[str, variables.Param],
-        Dict[str, variables.AssignedParam],
-        Dict[str, variables.SubscriptUse],
-        Dict[str, numpy.ndarray],
-        str,
-    ]:
-        if self.parameter_variables:
-            raise Exception("Compiler.compile() may be invoked only once per instance. Create a new instance to recompile.")
+            codegen_visitor = codegen_backends.EvaluateDensityVisitor(self.symbol_table, self._get_primary_symbol_from_statement(top_expr))
+            top_expr.accept(codegen_visitor)
```
bbbales2, line 696: I want to replace all the visitor stuff with:

```python
def evaluate_density_visitor(base_expr):
  for expr in base_expr:
    match expr:
      case ast.Subscript():
        ...
      case ast.Param():
        ...
```

If we do this we can delete all the `accept` functions from the ast and there's less function boilerplate on the codegen_backend part.

bbbales2 (commented):
I like the symbol table though I want to make sure and not oversimplify the subscripts. I only loosely followed the logic so I'd like to have follow up call there to discuss.

The visitor pattern on the codegen stuff seems a bit verbose, but I like how now that we have symbol table + ast we can just hand pieces of those off to different codegen units and do fancy things (which is nice). I left a proposal in the comments to simplify.

rat/compiler.py
Dashadower: The reason for this is because different backends can have different behavior for some nodes.
For example, if a backend subclass is identical to the evaluate density visitor except for a single node, we'd have to rewrite all the match cases.

rat/compiler.py
```diff
@@ -133,362 +384,190 @@ def _identify_primary_symbols(self):
...
-                                raise CompileError(
-                                    msg, self.model_code_string, primary_symbol.line_index, primary_symbol.column_index
-                                ) from e
-
-                            self.parameter_base_df[primary_key].columns = primary_df.columns
+                            else:
+                                for index, subscript_column in enumerate(primary_variable_record.subscript_alias):
+                                    if subscript_column not in self.data_df_columns:
+                                        # If there's a subscript name that's not present in the input dataframe, this means
+                                        # it's a subscript "alias", which points to the "true" subscripts that's already on
```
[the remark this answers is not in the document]
Dashadower: Yes, it's a rename. "aliases" are the column names of the base dataframe of that parameter. Looking back, it does sound confusing, but I thought them as aliases since your're renaming a set of input dataframe columns in to a single name.

rat/compiler.py
```diff
@@ -133,362 +384,190 @@ def _identify_primary_symbols(self):
...
-        for i, top_expr in enumerate(self.expr_tree_list):
-            primary_symbol = self._get_primary_symbol_from_statement(top_expr)
-
-            primary_key = primary_symbol.get_key()
-
-            # There aren't any restrictions on primary/secondary data
-            if isinstance(primary_symbol, ops.Data):
-                continue
+        for top_expr in self.expr_tree_list:
+            if isinstance(top_expr, ast.Distr):
```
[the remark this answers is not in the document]
Dashadower: Aye lol I keep forgetting about them

rat/compiler.py
```diff
@@ -7,82 +7,333 @@
...
+                raise CompileError(error_msg)
+
+            # Data variables don't need a base df; the input df is its base dataframe.
+            if record.variable_type == VariableType.DATA:
+                continue
+
+            if record.subscript_length > 0:
+                base_df = pd.DataFrame()
+
+                # Since the subscripts in the symbol table are all resolved to denote columns in the input dataframe,
```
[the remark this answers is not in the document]
Dashadower: Oh interesting. I though it still would be possible to access a subset of the base dataframe with columns from the input dataframe.
So to clarify, after a variable has been renamed(alias in this PR), it can only be accessed by its new names?

rat/compiler.py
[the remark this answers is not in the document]
Dashadower: Yup, the problem was with differentiating param and assigned param. An assigned param gets used on the rhs before it gets assigned as lhs.

rat/compiler.py
```diff
@@ -7,82 +7,333 @@
...
 
-        # these dataframes that go into variables.Subscript
-        self.data_base_df: Dict[str, pandas.DataFrame] = {}
+    def get_unique_number(self):
+        self._unique_number += 1
+        return self._unique_number
 
-        # this is the dictionary that keeps track of what names the parameter subscripts were defined as
-        self.parameter_subscript_names: Dict[str, List[Set[str]]] = {}
+    def upsert(
```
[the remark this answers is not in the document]
Dashadower: lol I'll happily take zoomer recommendations

rat/compiler.py
```diff
@@ -498,207 +577,155 @@ def _prepare_for_scans(self):
...
 
-                            shifts = symbol.subscript.shifts
+                            shifts = [shift.value for shift in symbol.subscript.shifts]
 
-                            if sum(shift is not None for shift in shifts) != 1:
+                            if sum(shift != 0 for shift in shifts) != 1:
                                 msg = "Exactly one (no more, no less) subscripts can be shifted in a recursively assigned variable"
                                 raise CompileError(msg, self.model_code_string, symbol.line_index, symbol.column_index)
 
-                            if shifts[-1] is None:
```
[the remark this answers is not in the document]
Dashadower: I changed the "rightmost subscript must have positive shift" condition for assignment into "a single subscript must have a positive shift".

rat/compiler.py
```diff
@@ -133,362 +384,190 @@ def _identify_primary_symbols(self):
...
+                        if primary_variable_record.subscript_length > 0:
+                            # these are the allowed subscript names in scope
 
-            for j in range(i + 1, N_lines):
-                for secondary_symbol in ops.search_tree(self.expr_tree_list[j], ops.Param):
-                    secondary_key = secondary_symbol.get_key()
+                            if primary_variable.subscript:
+                                for index, subscript_column in enumerate(primary_variable.subscript.names):
+                                    if subscript_column.name not in self.data_df_columns:
+                                        subscript_aliases[subscript_column.name] = {
```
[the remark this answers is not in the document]
Dashadower: Okay. Aliases are almost the same as renames in the current compiler. Take the simple baseball model:
```
score_diff ~ normal(skill[home_team] - skill[away_team], sigma);
skill[team] ~ normal(0.0, tau);
```
Previously, for parameter `skill`, we'd identify that the base_df should consist of input columns `home_team` and `away_team`, renaming its generated base_df's column name to `team`.

In the new compiler, the subscript alias `skill` denotes the set `{home_team, away_team}`, which is also used as its base_df column name. We still do the renaming - but just keep track of the "true" input dataframe column names that the renamed(alias) column name is created from.

rat/compiler.py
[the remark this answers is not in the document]
bbbales2: Well, the names aren't really used like that so much. Variables are subscripted by subscript names that belong to the primary dataframe.

Like:
```
y' ~ normal(mu[group], sigma);
```

`group` as a name belongs to the dataframe associated with y. `mu` builds itself a dataframe and gives it the name group, but the group name here is still associated with y.

rat/compiler.py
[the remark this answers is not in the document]
bbbales2: Aaah, k, I want to go ahead and materialize the intermediate dataframes for all the variables (and have the subscripts appear so that they are not represented as an expression + the input dataframe).

Reasons being:
1. It's simpler in my head (if it isn't, tell meeeee)
2. I want multiple input dataframes
3. With multiple input dataframes there will be joins and complicated transformations and I don't think the expression + input tracking will be easy

rat/compiler.py
bbbales2: Mmm I see. I do not like the visitor pattern, Sam I Am. Hmmm hmmmm hmmmm, is there anyway around this? Otherwise I am going to think about how to attack it from a do-we-need-it-perspective lol.

bbbales2: If there's no way around this and we need it then I'll be convinced lol, maybe I just gotta stare at it more.

bbbales2: > if a backend subclass is identical to the evaluate density visitor except for a single node, we'd have to rewrite all the match cases.

Okay so staring at this code we have EvaluateDensityVisitor and TransformedparametersVisitor. In each of these cases, the member functions have special overrides for certain things, like `visit_Param`, and `visit_Data`.

As we're parsing an expression tree, because of how the accept/visit logic works, if we hit a Param or a Data, we'll call our own special stuff before passing off to the more generic functions up the line (in BaseVisitor).

What if instead we did something like this:

```python
class EvaluateDensityVisitor(BaseVisitor):
  def my_data_logic(self, expr):
    pass

  def my_param_logic(self, expr):
    pass

  def generate(base_expr):
    for expr in base_expr:
      match expr:
        case ast.Data():
          self.my_data_logic(expr)
        case ast.Param():
          self.my_param_logic(expr)
        case _:
          self.generic_logic(expr) # This is implemented in BaseVisitor
```

bbbales2: Shit, I guess generic logic needs to somehow call back here again?

bbbales2: Noooooooo

rat/codegen_backends.py
```diff
@@ -33,187 +33,157 @@ def __init__(self, symbol_table: "compiler_rewrite.SymbolTable", primary_variabl
...
-    def visit_InverseLogit(self, invlogit_node: ast.InverseLogit, *args, **kwargs):
-        self.expression_string += "jax.scipy.special.expit("
-        invlogit_node.subexpr.accept(self, *args, **kwargs)
-        self.expression_string += ")"
-
-
-class EvaluateDensityVisitor(BaseVisitor):
+    def generate(self, ast_node: ast.Expr):
+        match ast_node:
+            case ast.IntegerConstant():
```
Dashadower, line 38: Can you check if your IDE picks up the node attributes(`subexpr, value, lhs, etc.`) within each case statement? Pycharm still identifies them as `Expr`, which is making me uncomfortable.

bbbales2: Yeah vscode is catching the types correctly. Maybe cause match is a 3.10 feature it'll just take a second till all the stuff is supported.

rat/codegen_backends.py
```diff
@@ -0,0 +1,302 @@
...
+                            )
+
+                            shift = [x for x in subscript_shifts if x != 0][0]
+                            self.expression_string += f"carry{shift}"
+                        else:
+                            self.expression_string += f"parameters['{param_key}']"
+                    elif param_key != self.lhs_key:
+                        self.expression_string += f"parameters['{param_key}']"
+                        self.expression_string += "["
+                        self.generate(ast_node.subscript, param_key)
```
bbbales2, line 246: I pass param_key as an argument instead of using the member variable. I'm not sure if it is that much of an improvement because I didn't change is_rhs.

tests/integration/test_time_series.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 import time
 
-from rat import ops
+from rat import ast
```
bbbales2, line 8: The time series tests aren't passing. I suspect something is broken with recursive assignment.

rat/codegen_backends.py
bbbales2: Also the cleanup looks great.

rat/codegen_backends.py
Dashadower: Thanks!

rat/codegen_backends.py
Dashadower: Hmm yeah I kept it as member variable to avoid `*args` hell but either seems fine.

tests/integration/test_time_series.py
Dashadower: Yeah this is bothering me alot lol. I manually checked the generated subscript indices and they match the original compiler. Still poking around to see where it's coming from.

rat/compiler.py
```diff
@@ -1,88 +1,51 @@
...
 
-class CompileError(Exception):
-    def __init__(self, message, code_string: str, line_num: int = -1, column_num: int = -1):
-        code_string = code_string.split("\n")[line_num] if code_string else ""
-        exception_message = f"An error occured while compiling the following line({line_num}:{column_num}):\n{code_string}\n{' ' * column_num + '^'}\n{message}"
-        super().__init__(exception_message)
+from . import ast
+from . import codegen_backends
+from .exceptions import CompileError
+from .symbol_table import SymbolTable, VariableType
```
Dashadower, line 14: I think these two modules are missing

rat/compiler.py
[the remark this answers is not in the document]
bbbales2: Ah, the reason it's the rightmost is performance. To do the scans we have to save in the rolling buffer so the previous element is in fact the previous element in that buffer. If we allow other positions then the rolling buffer logic is slower (and also not sure it would work with ragged array type stuff)

rat/codegen_backends.py
bbbales2: Okay I think the rule in a situation like this is solve the problem one way. I think that is_rhs is better implemented in the state. I'll revert this change.

tests/integration/test_time_series.py
bbbales2: Woof, probably some really fringe assumption I'm making with sorting or something lol.

rat/compiler.py
bbbales2: Ah, uploading now.

rat/codegen_backends.py
bbbales2: Reverted

rat/compiler.py
bbbales2: We made this change

rat/symbol_table.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class VariableType(Enum):
+    DATA = 1
+    PARAM = 2
+    ASSIGNED_PARAM = 3
+
+
+@dataclass()
+class TableRecord:
```
bbbales2, line 18: Actually question, if this is the symbol table, are the symbols the keys or the values?

What I've been saying is that the symbols are the pieces of the AST. In this case, there can be multiple symbols (for instance, parts of the AST in different statements) that point to the same values in the symbol table.

Dashadower: Right. I'd say it would be more appropriate to call it a "Variable table" instead. The SymbolTable keeps a single record for each variable, that contains any information that's necessary to construct the base dataframes and other subscripts.

Dashadower: So you can have multiple Data AST nodes in the expression tree named `y`, but they share the same record within the table.

tests/integration/test_time_series.py
Dashadower: I found a bug with setting pad_needed which resolved an error; two more to go!

Dashadower: I found the culprit! It was calculating `first_in_group_indicator` based on the joined dataframe instead of the variable's base_df.

rat/compiler.py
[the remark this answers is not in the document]
Dashadower: Ohhh I see. I'll revert that back

rat/compiler.py
```diff
@@ -7,82 +7,333 @@
...
+            record.constraint_lower = max(record.constraint_lower, constraint_lower)
+            record.constraint_upper = min(record.constraint_upper, constraint_upper)
 
-    def _get_unique_identifier(self) -> str:
+        else:
+            # Insert if new
+            self.symbol_dict[variable_name] = TableRecord(
+                variable_name,
+                variable_type,
+                subscripts=subscripts if subscripts else set(),
```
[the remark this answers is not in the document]
Dashadower: `subscripts` type is `Set[Tuple[str]]` so I think it makes sense to initialize to an empty set.

tests/integration/test_time_series.py
Dashadower: These integration tests are absolutely awesome <3

rat/compiler.py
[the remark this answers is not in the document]
bbbales2: Oh good point

tests/integration/test_time_series.py
bbbales2: Hell yeah hell yeah good sleuthing!

I've been really happy with what integration tests we have too. In that spirit, I'm pretty happy not going to a finer level of tests yet. So many of the details are changing so rapidly that I don't think going lower than this would be great.

rat/symbol_table.py
bbbales2: Nice nice, so symbols from the AST and then these are still Variables.

Dashadower:
I think this would be the end of changes for me. Feel free to add/revise and merge.

bbbales2 (commented):
Good!

bbbales2 (approved):
Good take #2!

PR #70, opened 2022-03-21
Updating docs for city rat

bbbales2:


bbbales2:
This has:
1. Updated terminology to reflect changes in how compiling works
2. Info on understanding prime syntax
3. Info on understanding recursive assignment

rat/__init__.py
```diff
@@ -235,106 +254,96 @@
...
-z[school] ~ normal(0, 1);
-mu ~ normal(0, 5);
-tau<lower = 0.0> ~ log_normal(0, 1);
-```
+This works because:
+1. Rat statements are executed across rows of the primary dataframe
+2. The primary dataframe is sorted in ascending order according to its subscripts
+3. The primary dataframe is already defined -- there is not problem of terminating
+this recursive statement
+4. Out-of-bounds accesses are mapped to zero (they do not throw errors)
```
Dashadower, line 302: It may be handy to emphasize that out of bound accesses are mapped to zero.

Dashadower (approved):
Looks good.

PR #73, opened 2022-04-10
Refactoring the symbol table

bbbales2:
Should be good to go.

Changes are:
1. Renamed symbol_table to variable_table (since symbols are the nodes in the AST)
2. Changed how iterating through the AST works
3. The base dataframe for parameter variables are computed on the fly


rat/variable_table.py
```diff
@@ -0,0 +1,242 @@
...
+        Add rows to the base_df
+        """
+        if self.base_df is None:
+            combined_df = new_rows_df
+        else:
+            combined_df = pandas.DataFrame(numpy.concatenate([self.base_df.values, new_rows_df.values]), columns=self.base_df.columns)
+
+        self.base_df = combined_df.drop_duplicates().sort_values(list(combined_df.columns)).reset_index(drop=True)
+
+    def set_constraints(self, constraint_lower: float, constraint_upper: float):
```
Dashadower, line 78: I think docstrings for the record methods should help remembering their uses in the future.

rat/variable_table.py
```diff
@@ -0,0 +1,242 @@
...
+        for column, shift in zip(shift_subscripts, shift_values):
+            shifted_column = grouped_df[column].shift(shift).reset_index(drop=True)
+            primary_base_df[column] = shifted_column
+
+        target_base_df = target_variable.base_df.copy()
+        target_base_df.columns = list(target_variable_subscripts)
+
+        target_base_df["__in_dataframe_index"] = pandas.Series(range(target_base_df.shape[0]), dtype=pandas.Int64Dtype())
+
+        key_name = f"subscript__{self.get_unique_number()}"
```
Dashadower, line 217: Looking at this made me remember something: do you think it would be worth "reusing" generated index arrays? Right now it generates a new index array every time `get_subscript_key` is called, but it can be changed so that if we already created indices that share the same primary and target variables, we return the name of the previously calculated array. 

Of course, that's added complexity to an already complicated function and I'm dubious if it's worth doing.

rat/variable_table.py
```diff
@@ -0,0 +1,242 @@
...
+        if self.constraint_lower != float("-inf") or self.constraint_upper != float("inf"):
+            if self.constraint_lower != constraint_lower or self.constraint_upper != constraint_upper:
+                raise AttributeError("Internal compiler error: Once changed from defaults, constraints must match")
+        else:
+            self.constraint_lower = constraint_lower
+            self.constraint_upper = constraint_upper
+
+
+class VariableTable:
+    symbol_dict: Dict[str, VariableRecord]
```
Dashadower, line 88: This should be changed to variable_dict for consistency.

Dashadower (approved):
Looks good!

rat/variable_table.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def set_subscript_names(self, subscript_rename: Tuple[str]):
+        if self.subscripts_rename is None:
+            self.subscripts_rename = subscript_rename
+            self.base_df.columns = self.subscripts
+        else:
+            if subscript_rename != self.subscripts_rename:
+                raise AttributeError("Internal compiler error: If there are multiple renames, the rename values must match")
+
+    def add_rows_from_dataframe(self, new_rows_df: pandas.DataFrame):
```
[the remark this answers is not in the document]
bbbales2: Yeah -- the VariableTable itself is acting more like a dictionary now, and then each element has its own update functions. I guess this is a more object oriented approach?

rat/variable_table.py
bbbales2: k will add

rat/variable_table.py
bbbales2: Hmm I guess it's possible but yeah I favor simplicity here until performance is a problem.

rat/variable_table.py
```diff
@@ -0,0 +1,242 @@
...
+                base_df = None
+
+            # Insert if new
+            self.symbol_dict[variable_name] = VariableRecord(
+                variable_name,
+                variable_type,
+                base_df,
+            )
+
+    def __contains__(self, name: str) -> bool:
```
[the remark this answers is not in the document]
bbbales2: Yeah I hope it's not too confusing from the outside (this thing looks like a dict but it isn't a dict, etc.)

rat/variable_table.py
bbbales2: Good catch -- changed

PR #77, opened 2022-04-16
Getting multiple dataframes working

bbbales2:
Nearly got stuff working. Hit a snag though.


rat/variable_table.py
```diff
@@ -233,22 +265,32 @@ def get_subscript_key(
             shifted_column = grouped_df[column].shift(shift).reset_index(drop=True)
             primary_base_df[column] = shifted_column
 
-        target_base_df = target_variable.base_df.copy()
-        target_base_df.columns = list(target_variable_subscripts)
+        target_base_df = target_variable.base_df[list(target_variable_subscripts)].copy()
```
bbbales2, line 268: The problem is here -- for parameters, subscript position matters. For how I envisioned data working, name matters. This is rather awkward.

My guess is long-term (maybe immediately after this PR or even in this PR) we should change how parameter names work so that renaming is no longer a thing. Something like:

```
score_diff' ~ normal(skills[team = home_team, year] - skills[team = away_team, year], sigma);
```

In this way the parameter names are sorta known always and there is no rename logic and position doesn't matter (and this matches the data behavior).

Dashadower: Hmm. I wonder whether reviving the subscript information on the variable table may help in solving this.
I think it still would be possible to keep track of the mapping from renamed subscript to original columns names internally with having the user specify them. Couldn't we be able to use that map for keeping the original names?

bbbales2: > I think it still would be possible to keep track of the mapping from renamed subscript to original columns names internally with having the user specify them. Couldn't we be able to use that map for keeping the original names?

We can certainly keep mappings from things to things but I just don't think it's useful. The rename (whatever form it takes) is very desirable in my opinion -- the original names aren't important.

If the original names were useful, then we'd not be combining them in a single variable. We'd do something like this:

```
score_diff' ~ normal(skills_home[home_team, year] - skills_away[away_team, year], sigma);
```

I could easily be misunderstanding tho so feel free to say so if I am.

Dashadower: So in this code:
```
score_diff' ~ normal(skills[home_team, year] - skills[away_team, year], sigma);
```
from what I understood, the problem you mentioned is in a case where `home_team` comes from dataframe A and `away_team` comes from dataframe B, and you use the parameter like so:
```
skills[team, year] ~ normal(0, 1);
```
since now the compiler can't figure out which columns should team be created from. Is that correct?

bbbales2: Aaah, in this specific case `home_team` and `away_team` come from the primary dataframe (the dataframe that score_diff comes from).

The reason I used that as an example was not that there were multiple dataframes, but as a different syntax for the renaming such that we could eliminate the importance of subscript position for parameters.

Dashadower: Oh okay I understood. I personally prefer to keep the current way of inferring renames from positions primarily due to it being simpler to write.

>For how I envisioned data working, name matters.
Can you explain how potentially positional renaming might be an issue here?

bbbales2: The logic for resolving the appropriate dataframe to look at for a variable is:

```
1. Variables with names that match columns in an *input dataframe* take that input dataframe
2. If variable names match columns in multiple input dataframes, it is an error
3. Otherwise, variables references are associated with parameters. The dataframe of the parameter is the minimum
```

In all cases, the subscript names themselves come from the primary dataframe (this could change but seems convenient for now).

bbbales2: Oh oh

bbbales2: Dangit my reply there with the rules -- that wasn't a reply to your message but an extra thing for the previous message whoops.

bbbales2: [Here](https://github.com/bbbales2/regressions/pull/77/files#diff-bbc8df5e38aac787903918b90f874a7b549d85962525bd8c5de7e0ba90490b83R275) is where the problem sticks out.

What this means is, if you want to join in a dataframe attached to data, you just list all the subscripts you wanna join by (the actual column joined is the one named). This is because the way data is joined the subscript names should be in both the primary and target dataframes.

If you want to join a dataframe attached to a parameter, you need to get parameter order right -- the names are there to reference the primary dataframe and the positions are used to reference the target dataframe columns.

It seems simpler that in both cases the names didn't matter and the names were expected to be the same in both tables -- this would require some inline renaming like `variable[name_in_target_table = name_from_primary_table]` I think.

bbbales2: Another place the difference shows up is on error checking in the compiler ([here](https://github.com/bbbales2/regressions/pull/77/files#diff-1e054d274852007ce49eb6ff658456b3b605510c572966506c6d38d11fcf6b0cR195))

Dashadower: Then wouldn't knowing which subscripts names(columns) from the input dataframe belong to which position be useful in resolving the new positions? For example, `skills` would have `[["home_team", "away_team"], ["year"]]` as its information on column order, and hence we should be able to match them back.

The renames would still be there - it's the column names of the dataframes associated to each parameter.

Dashadower:
```
score_diff' ~ normal(skills[home_team, year] - skills[away_team, year], sigma);
skills[team, year]' ~ normal(mean[team], 1.0);
```
joining param and param - no problem, we can just use the renames
```
some_other_score_diff' ~ normal(skills[team, year] - skills[team, year], sigma);
```
`team` not in input dataframe. But no problem, we know index 0 of skills parameter resorts to `home_team` and `away_team`, so reference that.

Dashadower: (the second code snippet doesn't really make sense model-wise but I think it conveys the point joining data and an already established parameter.)

Dashadower: actually,
```
some_other_score_diff ~ normal(skills[team, year]', sigma);
```
I think this should be correct

bbbales2: I'm nervous about trying to keep track of any information in a variable about a dataframe that it is not associated with.

I think the subscript AST is the right place to hold information about the relationship between these things, and that information is just, one dataframe has a thing X, and another will have the thing Y, and for the purposes of the calculation here these should be linked.

Dashadower: I wouldn't call it linking. More like we're tracing which subscript columns came from which columns from the input dataframe. All subscripts should be traceable to the input dataframes so that we know what columns to index when joining with data.

It's not easy to encode this info into the AST since we can't trace  the subscript creation until we actually traverse the AST and do the subscript resolving process.

Dashadower: Since the columns in the input dataframe are always there in place, and all subscript columns are essentially joins of columns from the input dataframe, IMO it's the easiest way to unify what would be used as the basis of joins that involve the input dataframe.

bbbales2: > It's not easy to encode this info into the AST

Yeah this is how I feel about this. We don't have an AST to achieve:

> All subscripts should be traceable to the input dataframes

What I mean is specifically easy to encode in the AST is just the local relationship between two variables in one statement -- not something from any of the potential input dataframes to any other input dataframe (or one of the constructed parameter dataframes).

> and all subscript columns are essentially joins of columns from the input dataframe

Not quite -- there's drop duplicates and concatenation going on as well, and they're build over a sequence of these operations.

The only advantage I see to tracing this information is maybe debug information, but I also don't know how to use that offhand.

bbbales2: I guess the concatenation and drop duplicates could all go at the end. But I still don't see an advantage of tracking this information.

Dashadower: > Not quite -- there's drop duplicates and concatenation going on as well, and they're built over a sequence of these operations.
Yeah you're right we can't just blindly recreate the dataframe with the original columns.

Dashadower: well shit lol maybe the renames are the only way?

Dashadower: damn... I wish there would be some way to automatically sort out the different names under hood

bbbales2:
@Dashadower this might be good, but I've forgotten lol. I see tests for the multiple dataframe stuff and I see docs, so I'm just gonna merge (I thought we merged this already)

PR #78, opened 2022-05-07
Adding more context to AST about source file

bbbales2:
I largely replaced line_index, col_index, and passing around the source file with a Position and Range class which are meant to signify position in text file and range in text file.

All tokens from the parser and expressions have ranges now. It's possible for an expression to have None as a range, which is annoying, but indicates the expression didn't come from source (we injected it). I'm not sure what to do about that but it doesn't seem great.

I added all this so it'd be easier to get range info back from language server.


bbbales2:
I'm gonna close this for now. I think the refactor here may as well go with all the other language stuff.

bbbales2:
Actually wait nevermind lol. This is probably better to merge as is even if more changes coming here. Will defer decision to later.

rat/exceptions.py
```diff
@@ -1,20 +1,40 @@
+from .position_and_range import Range
+
+# TODO: These four errors are basically all the same thing. Probably be possible to simplify them somehow
+
+
+class TokenizeError(Exception):
+    def __init__(self, message: str, range: Range):
+        code_string = range.document.split("\n")[range.start.line]
+        pointer_string = " " * range.start.col + "^" + "~" * max(0, range.end.col - range.start.col - 1)
```
Dashadower, line 9: The range indicator is pretty cool. How would this work on multiline errors?

Dashadower (approved):
LGTM

rat/exceptions.py
bbbales2: It'll do a bit of awkwardness :D. After looking at how the syntax highlighting works, I'm gonna change how these ranges are built to just be key tokens.

In the PR as is, the range for the `log` expression in `log(x + 1)` would be the full `log(x + 1)` text -- I want it to just be `log` now.

bbbales2: (and there's special multiline logic in the language server specification we can use if we ever wanna get there)

bbbales2: (and actually I'm not sure how ranges move between expressions -- anyway, this is just putting a bulk of edits here and then I'll change them in the follow up PR with the editor stuff)

PR #87, opened 2022-05-14
IfElse and recursion bugfix

### LINKED ISSUE #85: Issue with recursive assignment
This should compile and run but doesn't:

```python
def test_ifelse_recursive_assignment():
    data_df = pandas.read_csv(os.path.join(test_dir, "eight_schools.csv"))

    data_df = pandas.DataFrame({
        "school" : [1, 2, 3, 4],
        "offset" : [10.0, 0.0, 0.0, 0.0]
    })

    model_string = """
    # This first line more or less shouldn't matter
    offset' ~ normal(mu[school], 1000.0);
    mu[school]' = ifelse(
        school == 1,
        10.0,
        mu[shift(school, 1)] + 1.0
    ) + epsilon[school];
    epsilon[school] ~ normal(0.0, 0.1);
    """

    model = Model(data_df, model_string)
    fit = model.optimize()

    mu_df = fit.draws("mu")
```

Changing `school == 1` to `school[school] == 1` does the trick.

Dashadower:
Fixes #85 

The scan function works elementwise. In `TransformedParametersCodeGenerator` of codegen, we had a class attribute `self.at_rhs` which indicated whether we were generating code for scan(), which notified the codegen of `Subscript` to append `[index]` to parameters, which indexes parameters arrays. The same needed to be done for data arrays on the RHS, which was not being done.

### DIFF
--- a/rat/codegen_backends.py
+++ b/rat/codegen_backends.py
@@ -257,20 +257,31 @@ class TransformedParametersCodeGenerator(EvaluateDensityCodeGenerator):
         super().__init__(*args, **kwargs)
         self.lhs_used_in_rhs = False
         self.lhs_key = ""
-        self.at_rhs = False
+        self.at_rhs_of_scan = False
+        # at_rhs_of_scan is used to indicate whether we are generating code for the RHS of a jax.lax.scan().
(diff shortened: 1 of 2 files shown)


bbbales2 (approved):
Looks good!

Issue #88, opened 2022-05-14
Error when subtracting on left hand side of distribution


bbbales2:
```
rat.exceptions.ParseError: An error occured while parsing the expression at (0:7):
hscore - ascore ~ normal(skill[hteam] - skill[ateam], sigma);
       ^
Unknown operator '-' in statement
```

bbbales2:
Resolved by #92

PR #92, opened 2022-05-22
Allow expressions in LHS of sampling statement

Dashadower:
Addresses #88. Uses precedence values to accomplish this.


bbbales2 (approved):
Looks good to me. I added a test.

Issue #89, opened 2022-05-14
Not checking that columns actually exist in source dataframes


bbbales2:
`score_diff` didn't actually exist in input dataframe. I got this error:

```
rat.exceptions.CompileError: An error occurred while compiling code at (0:27):
score_diff' ~ normal(skill[hteam] - skill[ateam], sigma);
                           ^~~~~
Subscript hteam not found in dataframe of primary variable score_diff
```

Dashadower:
Does `hteam` exist?

Dashadower:
If you can link model code and data, I can take a closer look.

bbbales2:
I added a test in a PR over [here](https://github.com/bbbales2/regressions/pull/93)

Dashadower:
Closing since error was from a different source. But we should revisit and perhaps add a check for parameters without priors set.

PR #93, opened 2022-05-22
Fix issue 89

bbbales2:
This isn't actually a fix, but it's a test for #89. I think issue there is more or less the wrong error message is being thrown, so I added a test here that expects an error message about the primary variable not being found.


Dashadower (approved):
The compiler was correct :)

PR #95, opened 2022-06-12
Working on iteratively inferring parameter domains

bbbales2:


rat/grammar.ebnf
```diff
@@ -0,0 +1,56 @@
...
+    | function_call
+    | subexpression
+    | variable
+    | literal
+;
+
+ifelse::IfElse = 'ifelse' ~ '(' predicate:expression ',' left:expression ',' right:expression ')';
+function_call::FunctionCall = name:identifier '(' [ arglist:arglist ] ')';
+variable::Variable = name:identifier [ constraints:constraints ] [ '[' arglist:arglist ']' ] [ prime:"'" ];
+constraints::Constraints = '<' ~ left:constraint ~ {',' right:constraint} '>';
```
Dashadower, line 43:
```
{',' right:constraint}
```
I think this should be:
```
[',' right:constraint]
```
Since we only allow at most 2 constraints.

rat/compiler2.py
```diff
@@ -0,0 +1,747 @@
...
+from rat import subscript_table
+
+
+@dataclass
+class StatementInfo:
+    statement: ast.Statement
+    primary: ast.Variable
+
+
+def _get_subscript_names(node: ast.Variable) -> List[str]:
```
Dashadower, line 30: This is confusing since we're not running this on statements but on variable nodes.

Dashadower: Triple nested walkers lol

rat/grammar.ebnf
bbbales2: Makes sense. Change awayyy

rat/grammar.ebnf
```diff
@@ -0,0 +1,56 @@
...
+    | ifelse
+    | function_call
+    | subexpression
+    | variable
+    | literal
+;
+
+ifelse::IfElse = 'ifelse' ~ '(' predicate:expression ',' left:expression ',' right:expression ')';
+function_call::FunctionCall = name:identifier '(' [ arglist:arglist ] ')';
+variable::Variable = name:identifier [ constraints:constraints ] [ '[' arglist:arglist ']' ] [ prime:"'" ];
```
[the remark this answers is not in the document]
bbbales2: Feel free to change it if it makes stuff more convenient.

There's apparently some polymorphism-kinda stuff exposed in Tatsu: https://tatsu.readthedocs.io/en/stable/models.html#model-class-hierarchies, but I didn't manage to get it to work when I tried it before. May be worth revisiting!

If it worked then maybe we can have cake and eat it?

rat/compiler2.py
bbbales2: Looool, yeah I went kinda nuts with the little walker things. I quite liked them, they seem very modular in a useful way. 

Like, even if sometimes stuff feels a little old-fashion (passing arguments through member variables), it also feels very self-contained and understandable.

Before (and still now) we have this giant compiler class that has a bunch of functions that all sorta come together and do work. With the walkers it seems like we could just put each walker in its own file in a folder called "Walkers" and then call them from other places and then the compiler code itself is smaller, which seems nice. Anyway, that could have its own drawbacks lol.

rat/grammar.ebnf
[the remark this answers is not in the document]
Dashadower: I decided to stick with the current approach for now.

Dashadower:
https://docs.google.com/document/d/1urfnRmeexmTS_rBSL7KlmNqON9f5dx26PyOnpSxhaHI/edit?usp=sharing

I'd like to try and find the domains with data flow analysis. I think having a separate control flow graph rather would be much easier to work with than an AST for this kind of thing. The only portions we have to codegen and execute are the conditionals and the trasfer functions.

Dashadower:
Are the prime symbols still required or can we get buy without them?

rat/codegen_backends.py
```diff
@@ -1,357 +1,269 @@
...
-                    self.expression_string += "\n"
+                    index_lookup = ""
+
+                return f"parameters['{node.name}']{index_lookup}"
 
+    def walk_Literal(self, node: ast.Literal):
+        return f"{node.value}"
+
+
+class CodeExecutor(NodeWalker):
```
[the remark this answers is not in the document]
bbbales2: CodeExecutor runs when Jax traces the transformed parameters and evaluating densities code.

Also be warned that transformed parameters has a required execution order but I don't think there are checks enforcing this anymore lol.

rat/codegen_backends.py
```diff
@@ -1,357 +1,269 @@
...
+            return_value = self.walk(node.right)
+            self.left_side_of_sampling = None
+            return return_value
+        else:
+            raise CompileError(f"{node.op} operator not supported in BaseCodeGenerator", node)
+
+    def walk_Logical(self, node: ast.Logical):
+        left = self.walk(node.left)
+        right = self.walk(node.right)
+        return eval(f"(left {node.op} right)")
```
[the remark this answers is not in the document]
bbbales2: Eval acts just like running code I think. It can see that left and right have been defined lol. It's a bit funky but works.

I use `eval(node.name)` for functions. This basically uses eval to do a string lookup by function-name for whatever it should be executing. Someone on Stackoverflow showed this.

We could probably do something less open-to-mistakes than this but it was easy to code up :P.

rat/codegen_backends.py
```diff
@@ -1,357 +1,269 @@
...
+
+class TraceExecutor(NodeWalker):
+    """
+    As long as shunt is true, values of expressions aren't returned but
+    instead shunted off into tuples
+    """
+
+    tracers: Dict[str, Tracer]
+    leaves: Dict[str, Any]
+    shunt: ContextStack[bool]
```
Dashadower, line 203: I don't understand how the exact behavior of `shunt` is related with respect to `first_level_values`. I see that shunt is used in variable subscripts to indicate that values should be evaluated and returned instead of being traced(recorded), but how is first_level_values related?

rat/variable_table.py
```diff
@@ -14,6 +16,104 @@ class VariableType(Enum):
...
+
+    def copy(self):
+        return Tracer(self.args_set.copy(), self.arg_types)
+
+
+K = TypeVar("K")  # TODO: Is this a useful way of typing this?
+V = TypeVar("V")
+
+
+class ValueTracer(Tracer, Generic[K, V]):
```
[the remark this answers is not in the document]
Dashadower: Or perhaps `ValueTracer` and `IndexTracer`

rat/variable_table.py
```diff
@@ -14,6 +16,104 @@ class VariableType(Enum):
...
+    def __repr__(self):
+        return f"F[{str(self.arg_types)} -> ?]"
+
+    def argument_length(self):
+        if self.arg_types:
+            return len(self.arg_types)
+        else:
+            raise Exception("Argument length called before argument length known")
+
+    def __call__(self, *args):
```
Dashadower, line 43: I'm wondering if `__call__` would be necessary given we aren't calling the tracers within `eval` statements. I'd prefer a method with an explicit name like `register(*args)` instead and invoke it.

rat/codegen_backends.py
```diff
@@ -1,357 +1,269 @@
...
+        return self.walk(node.left) if predicate else self.walk(node.right)
+
+    def walk_FunctionCall(self, node: ast.FunctionCall):
+        if node.arglist:
+            arglist = tuple(self.walk(arg) for arg in node.arglist)
+        else:
+            arglist = ()
+
+        if not self.shunt.peek():
+            return eval(node.name)(*arglist)
```
Dashadower, line 246: The math functions in rat are defined simply as `log`, `sin`, but looks like this would fail because instead it should be calling `jax.numpy.log` and so on.

Dashadower (commented):
I like the overall idea of building a structure of tracers which hold all the input/output values of all variables and unifies them with integer indices. The subscript table generation by walking through the AST was pretty neat and clever.

However, I think the code overall got *way* more complex than the previous compiler, which I suppose is due to the addition of a lot of state variables here and there making reading through the overall code flow quite hard without using the debugger often.

I we think should still be able to drop primary variable marking and support the indication of param/data variables through the input data. The idea is that the variable with the highest subscript count/length would be primary. If there are multiple variables with same number of subscripts, we'd be using the one that has already been traced.

I also think we can support replacing parameter values as data without much change? For example in the eightschools model, if I supplied an additional dataframe with columns `z, school` then I think the tracer will correctly read it and set `z` as data. I think the remaining portion is removing any statements where `z` is used as LHS and RHS contains no parameters.

Dashadower:
```
y[school]' ~ normal(theta[school], sigma[school]);
theta[school]' = mu + z[school] * tau;
z[school] ~ normal(0, 1);
mu ~ normal(0, 5);
tau<lower = 0.0> ~ log_normal(0, 1);
```

I'm surprised how it takes 3 trace iterations to finish tracing. I'm expecting just one iteration since there's no recursion stuff  going on.

rat/compiler2.py
```diff
@@ -0,0 +1,737 @@
...
+                primary_variable = self.variable_table[primary_ast_variable.name]
+
+                for row in primary_variable.itertuples():
+                    executor = TraceExecutor(tracers, row._asdict())
+                    executor.walk(statement_info.statement)
+
+            found_new_traces = False
+            for variable_name, tracer in tracers.items():
+                variable = self.variable_table[variable_name]
+                found_new_traces |= variable.ingest_new_trace(tracer)
```
Dashadower, line 381: Hmm. Updating the variable table at the end results in multiple tracing iterations needed to evaluate non-recursive programs. Updating the variable table while we're tracing will fix this, since then multiple dependent variables can be traced in a single trace, but then you'd need more logic and code to check if there are new traces(comparing the lengths of all tracer arguments before and after the trace iteration). Would this be worth?

rat/compiler2.py
```diff
@@ -0,0 +1,737 @@
...
+
+                def walk_Variable(self, node: ast.Variable):
+                    if node.name != self.primary_name:
+                        if node.arglist:
+                            variable = self.variable_table[node.name]
+                            subscript_names = _get_subscript_names(node)
+
+                            if variable.variable_type != VariableType.DATA and not variable.renamed and subscript_names is not None:
+                                try:
+                                    variable.suggest_names(subscript_names)
```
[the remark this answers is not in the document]
bbbales2: If it hasn't been renamed, yes it would try to use the given names and fail if there are multiple options that aren't equivalent.

rat/variable_table.py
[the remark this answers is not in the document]
bbbales2: Fair point. I'd like to somehow only have one tracer class. I'll defer a rename until I have a better idea how to refactor this and the VariableTable/VariableRecord classes.

rat/codegen_backends.py
bbbales2: It's awkward that so much happens here -- I think it wouldn't be bad to split this in two except then I'd be repeating myself for all but a couple bits of logic :/.

rat/variable_table.py
bbbales2: Hmm good point. This is from the codegen era but we have more control on our tracer calls now that we're not doing codegen. Will just add a note to code and save this for VariableTable refactor whenever that happens.

rat/codegen_backends.py
bbbales2: Good point. Yah boy will just sneak in a from jax.numpy import * for now.

rat/compiler2.py
bbbales2: Until we see a performance problem I lean no. I expect we'll see a performance problem eventually lol.

rat/codegen_backends.py
bbbales2: It's totally possible we do something clean here that isn't just importing namespaces and hoping they don't clobber each other. I feeling is we wait until we allow people to pass in custom user defined python functions and bind them to names in Rat to worry about this tho.

bbbales2:
> I we think should still be able to drop primary variable marking and support the indication of param/data variables through the input data.

Possibly. I'm not gonna complain if there's more automated logic here, even if we can never really get rid of the marking itself.

I've not been able to come up with syntax that is clear though. The thing I've been thinking is something like:

`expression "over" variable_name`

So:
```
y[n]' ~ normal(mu[group[n]], sigma)
```

Could be:
```
y[n] ~ normal(mu[group[n]], sigma) over y
```

> I also think we can support replacing parameter values as data without much change?

Possibly. I've been thinking about mixed data/parameter variables for missing data. It seems very do-able. At this point I'd like it more to unify code rather than actually handle missing data. I'm avoiding it till I know I need it tho.

bbbales2:
Just realized it's also very nice to know exactly what the primary variable is without the data. There's a fair bit of logic we can't do until we know the primary variable.

rat/model.py
```diff
@@ -1,272 +1,340 @@
...
-                raise Exception("Only one of model_string and parsed_lines can be non-None")
+    def walk_FunctionCall(self, node: ast.FunctionCall):
+        argument_list = []
 
-            parsed_lines = []
-            scanned_lines = Scanner(model_string).scan()
-            for scanned_line in scanned_lines:
-                parsed_lines.append(Parser(scanned_line, data_names, model_string).statement())
+        if self.left_side_of_sampling:
+            argument_list += [self.walk(self.left_side_of_sampling)]
```
Dashadower, line 61: Why is the LHS variable's values being added?

rat/model.py
```diff
@@ -1,272 +1,340 @@
...
-        compile_path: str = None,
-        overwrite: bool = False,
-    ):
-        """
-        Create a model from some data (`data`) and a model (specified as a string, `model_string`).
 
-        If compile_path is not None, then write the compiled model to the given path (will only overwrite
-        existing files if the overwrite flag is true)
+class CodeExecutor(NodeWalker):
+    arguments: Dict[ast.Variable, Any]
```
Dashadower, line 29: This should indicate they're subscript/data values. Arguments is confusing.

rat/model.py
```diff
@@ -1,272 +1,340 @@
...
+    parameters: Dict[str, Any]
```
Dashadower, line 30: I would also appreciate a more verbose name, such as `parameter_value_arrays`

rat/trace_table.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Dict, Iterator, Tuple
+
+
+@dataclass(frozen=True)
+class TraceRecord:
+    name: str
+    array: numpy.ndarray
+
+
+class TraceTable:
```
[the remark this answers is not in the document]
Dashadower: I've re-read the code and the trace table also seems like it keeps track of values for data variables.

Dashadower (commented):
Some Qs and suggestions.
Overall, calculating parameter values inplace is neat, but tbh kinda makes the code hard to follow through.

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
+    if len(candidates) == 0:
+        msg = f"No primary variable found on line (this means there are no candidate variables)"
+        raise CompileError(msg, statement)
+
+
+# 1. If the variable on the left appears also on the right, mark this
+#   statement to be code-generated with a scan and also make sure that
+#   the right hand side is shifted appropriately
+@dataclass
+class CreateAssignedVariablesWalker(RatWalker):
```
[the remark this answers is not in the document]
bbbales2: Yeah those are broken as of now. If statements then that, and it'll be a trick for the recursive things no matter how it's sliced I think.

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
-        # TODO: I'm not sure it's a good pattern to modify the AST in place
-        # Fold all shifts to constants
-        for top_expr in self.expr_tree_list:
-            for symbol in top_expr:
-                match symbol:
-                    case ast.Data() | ast.Param():
-                        if symbol.subscript is not None:
-                            integer_shifts, padding_needed = self._get_shifts_and_padding(symbol.subscript)
+        if node.arglist:
+            with self.in_control_flow.push(True):
```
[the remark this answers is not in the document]
bbbales2: Hmm, I guess the funny thing here is that we never push false. It's you go into control flow and you're there. I think this could just be replaced by a default-to-false bool. Does that cover your Q?

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
+                for arg in node.arglist:
+                    self.walk(arg)
 
-                            new_shift_expressions = []
-                            for shift_expression, integer_shift in zip(symbol.subscript.shifts, integer_shifts):
-                                new_shift_expressions.append(ast.IntegerConstant(value=integer_shift, range=shift_expression.range))
 
-                            symbol.subscript.shifts = tuple(new_shift_expressions)
+@dataclass
+class RenameWalker(RatWalker):
```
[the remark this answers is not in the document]
bbbales2: Sounds good, go for it

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
+@dataclass
+class InferSubscriptNameWalker(RatWalker):
+    variable_table: VariableTable
+    primary_name: str = None
+
+    def walk_Statement(self, node: ast.Statement):
+        primary_variable = get_primary_ast_variable(node)
+        self.primary_name = primary_variable.name
+
+    def walk_Variable(self, node: ast.Variable):
```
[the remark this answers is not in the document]
bbbales2: looool good call. I believe you're right. This code should run tho, so we should wire that back up.

rat/variable_table.py
```diff
@@ -1,344 +1,265 @@
...
-            if series.dtype == pandas.Int64Dtype():
-                return series.astype(int).to_numpy()
-            else:
-                return series.to_numpy()
-        else:
-            raise AttributeError("Internal compiler error: `to_numpy` can only be used if variable_type is data")
+    def buffers_are_different(self):
+        raise Exception("Internal error: Use a dynamic subclass instead")
 
+    def itertuples(self) -> Iterable[NamedTuple]:
```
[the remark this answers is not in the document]
bbbales2: Yah sure thing I just lazily copied the name from here: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.itertuples.html

We're also iterating over the value of the function when we have them which makes this kinda tricky (sometimes it's just domain -- sometimes it's domain and something else). Feel free to rename. The usage is a bit tricky. Not sure what to do.

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
+
+    def __init__(self, variable_table: VariableTable):
+        super().__init__()
+
+        self.variable_table = variable_table
+
+        for name in self.variable_table:
+            variable = self.variable_table[name]
+            if isinstance(variable, DynamicVariableRecord):
+                variable.swap_and_clear_write_buffer()
```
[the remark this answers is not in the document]
bbbales2: The variable domains are double buffered. You read from one while you write to the other and swap back and forth.

So first read from A and write to B, and then read from B and write to A, and then A to B again, etc.

At least that's the intention

rat/trace_table.py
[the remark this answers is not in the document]
bbbales2: Yeah the trace table zoops out everything that can be data into 1D numpy arrays that can be consumed by jax as easily as possible. There's some inefficiency for sure (literals get expanded to vectors). Currently it'll also record traces for things that are straight up unused.

The trace word now means something very different than it did previously.

The design here was basically, if we execute on the ast, then when jax hits a node that isn't autodiffed, it needs to be able to go somewhere and just grab the vector corresponding to that node. That's what tracetable does, though it's less efficient than it probly needs to be lol.

rat/model.py
bbbales2: This is analogous to the Stan conversion of:

```
y ~ normal(mu, sigma);
```

to

```
target += normal_lpdf(y, mu, sigma);
```

rat/model.py
bbbales2: Oh hmm you're right. `traced_values` or `data` or something seems better.

rat/model.py
bbbales2: We could split these by autodiff/non-autodiff or something? I kinda like parameters here. I think I used parameters just cause that's what I called it in other places and that's what it was in the codegen version before.

bbbales2:
> calculating parameter values inplace is neat

Is this specifically the DomainDiscoveryWalker or more broadly than that?

Dashadower:
> > calculating parameter values inplace is neat but tbh kinda makes the code hard to follow through
> 
> Specifically the DomainDiscoveryWalker or more broadly than that?

Yeah pretty much that class

rat/compiler.py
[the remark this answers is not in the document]
Dashadower: Yeah switching to just a bool seems better.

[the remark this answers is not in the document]
Dashadower: https://github.com/bbbales2/regressions/blob/37b9d2f8c7c4d63a59db9df1079d8fa08018eb0d/rat/codegen_backends.py#L53

Well it does pop the `True` value after it's done though. I actually think we can keep it as-is.

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
+        subscript_names = _get_subscript_names(node)
+
+        if node.arglist:
+            argument_count = len(node.arglist)
+        else:
+            argument_count = 0
+
+        if unallocated:
+            if self.in_control_flow.peek():
+                record = ConstantVariableRecord(node.name, argument_count)
```
Dashadower, line 158: Am I right that variables defined from data should be designated as `ConstantVariableRecord`? If so, this logic doesn't catch subscript variables. For example, in the eightschools code(with `eightschools.rat` and `eight_schools.csv` in the integration tests directory), `sigma` will be designated with `SampledVariableRecord`

rat/compiler.py
```diff
@@ -1,488 +1,439 @@
...
-        self.variable_table = VariableTable(self.data)
-
-        # Add entries to the symbol table for all data/params
-        for top_expr in self.expr_tree_list:
-            for symbol in top_expr:
+                        # It's okay if an argument_count == 0 something doesn't bind -- it (hopefully) will be provided by
+                        #  the primary variable instead
+                        return
+            else:
+                record = ConstantVariableRecord(node.name, argument_count)
```
Dashadower, line 170: Also marking here since this also handles `ConstantVariableRecord`

rat/model.py
bbbales2: Maybe the naming should be Compiletime vs. Runtime, or Constant vs. Dynamic (and we already use the constant/dynamic verbage on the variable table).

The data/parameter split is awkward here because transformed parameters are dynamic without being sampled parameters.

rat/compiler.py
bbbales2: Hmm, looking through ConstantVariableRecord.bind there's a lot of ways this can fail and many of these should be fatal. I think the only error we'd want to actually catch and ignore would be the stuff corresponding to:

```
try:
  data_name = get_dataframe_name_by_column_name(self.name, data)
except KeyError:
  raise KeyError(f"{self.name} not found in the input data")
```

Everything else I think should be fatal cause it corresponds to situations like, we found a matching variable in an input dataframe but the subscripts didn't match, or we had a problem with the subscripts, or there were duplicate rows in the input, etc.

`tests/integration/test_model_noncenter.py` works for me, so if it's failing for you and it's because sigma is trying to be a parameter, I'd change this error logic to only ignore the very specific behavior we want to ignore.

So maybe add a:
```
class VariableNotFoundInData(ValueError):
  pass
```

And then throw and catch only that very specific error (so that anything else will explode and get you more info)

Dashadower:
```
y[school]' ~ normal(theta[school], sigma[school]);
```
Looks like the problem occurs when this is replaced into:
```
y[school]' ~ normal(theta[school], sigma);
```
Are we rolling with mandating subscripts even for entirely row-based statements?

rat/model.py
Dashadower: Yup - I like the compile-time vs runtime separation. I'm not sure whether `CodeExecutor` would be the place to nuance this separation since this is just a generic executor class.

bbbales2: No need to worry if this is perfect place or not. If it's easier to read it's easier to read. Can always change the naming later.

rat/compiler.py
bbbales2: Yeah we're requiring subscripts. I forget where we discussed this, but I think this gets rid of any special-ness of data in a dataframe. Maybe long term we change our tune here (cuz it seems awkward to need to specify sigma if they're literally on the same row in a dataframe). This is simpler though, cause we don't need to worry about how data is arranged in a dataframe to make this work.

rat/grammar.ebnf
```diff
@@ -37,10 +37,10 @@ factor
     | literal
 ;
 
-ifelse::IfElse = 'ifelse' ~ '(' predicate:expression ',' left:expression ',' right:expression ')';
+ifelse::IfElse = 'ifelse' '(' ~ predicate:expression ',' left:expression ',' right:expression ')';
 function_call::FunctionCall = name:identifier '(' [ arglist:arglist ] ')';
 variable::Variable = name:identifier [ constraints:constraints ] [ '[' arglist:arglist ']' ] [ prime:"'" ];
-constraints::Constraints = '<' ~ left:constraint ~ [',' right:constraint] '>';
+constraints::Constraints = '<' left:constraint ~ [',' right:constraint] '>';
```
bbbales2, line 43: This was a bug that made it so that when the parser found:

`identifier <`, it thought that it had to match a constraint.

In reality that could also be `identifier < identifier`, or something else.

rat/compiler.py
```diff
@@ -329,15 +332,24 @@ def walk_Statement(self, node: ast.Statement):
...
+
+        dense_size = len(primary_variable)
+        for traced_node, trace in traces.items():
+            idxs, values = zip(*trace)
+
+            numpy_idxs = numpy.array(idxs)
+            numpy_values = numpy.array(values)
+
+            dense_values = numpy.zeros(dense_size, dtype=numpy_values.dtype)
+            dense_values[numpy_idxs] = numpy_values
```
bbbales2, line 350: This change makes it so that any time we trace something, we pad stuff out so we have arrays with length equal to the size of the domain of the primary variable.

Ifelses make it possible for variables get used only some of the time during tracing. vmapping requires that all input arrays be equal length. This is padding everything out to avoid vmap errors.

rat/grammar.ebnf
Dashadower: Good catch!

.github/workflows/unit_and_integration_tests.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       run: |
         pip install .
         pip install pytest
+	bash update_grammar.sh
```
Dashadower, line 22: Looks like an indent error

rat/__init__.py
```diff
@@ -101,7 +103,146 @@
...
+   1,  thomspon
+```
+
+```
+made' ~ bernoulli_logit(skill[player]);
+```
+
+In this case the syntax suggests that there is one `skill` parameter for each
+player somewhere. Rat treats unresolved parameters like this like functions and
+is in this way works similarly to Bean Machine.
```
Dashadower, line 180: I would eventually support saying parameters that are uninitialized with a prior should be illegal in rat.

rat/model.py
```diff
@@ -1,272 +1,349 @@
...
-        return jacobian_adjustment, self.compiled_model.transform_parameters(
-            self.device_data, self.device_subscript_indices, self.device_first_in_group_indicators, parameters
-        )
-
-    def _log_density(self, include_jacobian, unconstrained_parameter_vector):
-        # Evaluate model log density given model, data, subscripts and unconstrained parameters
-        jacobian_adjustment, parameters = self._constrain_and_transform_parameters(unconstrained_parameter_vector, pad=True)
-        target = self.compiled_model.evaluate_densities(self.device_data, self.device_subscript_indices, parameters)
-        return target + (jacobian_adjustment if include_jacobian else 0.0)
+class ExecuteException(AstException):
```
Dashadower, line 23: Why does this have to be outside `exceptions.py`?

rat/compiler.py
Dashadower: Hmm. Can there be a problem with having zeros where we shouldn't have them? What if we had something like `ifelse(x < 1e-5, 0, log(x))`?

Dashadower (commented):
I'm still looking through the files. It's taking a little bit since there are changes that were already part of old commits that we've gone through, and github diff viewer being super slow.

Dashadower (changes requested):
Code involving assignments are currently failing due to the reason I've specified in the comments.

rat/compiler.py
```diff
@@ -1,488 +1,571 @@
...
+            )
+            raise CompileError(msg, candidates[0])
+
+    if len(candidates) == 0:
+        msg = f"No primary variable found on line (this means there are no candidate variables)"
+        raise CompileError(msg, statement)
+
+
+@dataclass
+class AugmentDataVariableSubscriptsWalker(RatWalker):
```
Dashadower, line 124: I've added this walker in the compiler to handle data variables without subscripts. Please read the class docstring.

setup.py
```diff
@@ -24,4 +24,5 @@
     python_requires=">=3.10",
     install_requires=requirements,
     zip_safe=False,
+    package_data={"rat": ["grammar.ebnf"]},
```
Dashadower, line 27: This file wasn't being added when you installed the package, so I've manually stated the file to be included during installation.

Dashadower (commented):
I've made some new changes. I think you should be aware of them.

setup.py
bbbales2: Oh cool. Did this lead to an error? I do `pip install -e .` to develop locally so I miss stuff like this.

rat/grammar.ebnf
```diff
@@ -0,0 +1,56 @@
+@@grammar::Rat
+@@eol_comments :: /#.*?$/
+@@keyword :: let in end fun
+@@parseinfo :: True
+
+start::Program = { statements+:statement }+ $;
+
+statement::Statement = left:factor ~ op:('~' | '=') ~ right:expression ';';
```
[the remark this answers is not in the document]
bbbales2: Hrmm, I maybe just try promoting the left hand side to an expression? I think you hit a problem when the left-most-this is kinda recursively defined.

Like this would be a problem:
```
statement = left:statement '~' right:statement
```

I'm not really sure what the rules are but you can see that in the mini calc example: https://tatsu.readthedocs.io/en/stable/mini-tutorial.html

I think we can't assign to expressions tho, so this'd need to break out into one rule for `~` and one rule for `=`.

rat/compiler.py
bbbales2: > y ~ normal(mu, sigma);

This makes sense. My assumption at this point was subscripts were always required. (I just hadn't updated all the tests)

I see the index solution. Maybe there's an edge case if we implement scalar data here? I'm not sure. There'd be the other danger that another dataframe can't join in to this by index (and two things named index are not the same).

rat/model.py
```diff
@@ -1,272 +1,349 @@
...
+
+        primary_node = compiler.get_primary_ast_variable(node)
+        primary_name = primary_node.name
+        primary_variable = self.variable_table[primary_name]
+
+        if node.left.name != primary_name:
+            msg = f"Not Implemented Error: The left hand side of assignment must be the primary variable for now"
+            raise AstException("computing transformed parameters", msg, node)
+
+        dependency_finder = CodeExecutorDependencyFinder(self.trace_table)
```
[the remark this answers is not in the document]
bbbales2: We're filtering out assignments at the top of this function tho so this seems fine?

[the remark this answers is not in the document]
bbbales2: Hmm lol, I see.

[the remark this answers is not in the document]
bbbales2: I guess I just never tested this lol.

rat/model.py
```diff
@@ -1,272 +1,349 @@
...
-                                data_names.add(column)
-                        case _:
-                            raise Exception(f"Values of dictionary form of data must be pandas DataFrames, found {type(value)}")
-            case _:
-                raise Exception("Data must be a pandas DataFrame or a dictionary")
+    def walk_Statement(self, node: ast.Statement):
+        if node.op == "~":
+            self.walk(node.left)
+            self.walk(node.right)
+        else:
```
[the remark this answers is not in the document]
bbbales2: This is figuring out which nodes should come from the trace table. I believe it should work on assignments or sampling statements. I agree that we should do no check here and just run.

rat/compiler.py
```diff
@@ -1,488 +1,501 @@
...
+
+            self.variable_table[node.name] = record
+        else:
+            record = self.variable_table[node.name]
+
+            if isinstance(record, ConstantVariableRecord):
+                if argument_count == 0:
+                    if node.name not in self.primary_node_and_subscript_names:
+                        msg = (
+                            f"{node.name} is a constant variable and must be a primary variable subscript if it appears with no subscripts"
```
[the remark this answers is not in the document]
bbbales2: Oh, so don't actually traverse if statements here?

I think we'd still need the in_control_flow logic but breaking this into two seems fine.

bbbales2:
That's a bummer that the error strings aren't doing good. Any particular tests looking wacky? I thought I had the generic parts of the exception logic that handle the prints looking good. Maybe we're passing the wrong nodes to the error messages?

rat/compiler.py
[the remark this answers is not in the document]
Dashadower: We so need to traverse through if statements. But since it goes through the arglist of variables, it flags `in_control_flow` to true even for variables that aren't in predicates of ifelse.

rat/compiler.py
Dashadower: Ah okay. I was going over the tests and some of them weren't equipped with subscripts. Not sure though since adding a `n` column to the dataframe (like in eightschools) pretty does the same thing.

setup.py
Dashadower: I also have it installed in editable mode, but then realized tests ran on actions were giving that filenotfounderror. So I tried installing rat on a brand new venv without the editable flag, and saw that the ebnf file doesn't get copied over.

Dashadower:
> That's a bummer that the error strings aren't doing good. Any particular tests looking wacky? I thought I had the generic parts of the exception logic that handle the prints looking good. Maybe we're passing the wrong nodes to the error messages?

The error that's looking weird is the one with `CodeExecutorDependencyFinder`. It will say it doesn't support  the `=` operator, but the rat code string that also gets printed out was a sampling statement, with the line column caret pointing to EOL.

rat/grammar.ebnf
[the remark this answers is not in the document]
Dashadower: Yeah we do need to separate them.

rat/variable_table.py
```diff
@@ -1,344 +1,295 @@
...
-import pprint
-from typing import Dict, Set, Tuple, Iterable, Union
+from sortedcontainers import SortedDict, SortedSet
+
+K = TypeVar("K")
+V = TypeVar("V")
 
-from .exceptions import CompileError, MergeError
 
+def get_dataframe_name_by_variable_and_column_name(
```
Dashadower, line 11: I've updated this file so it supports multiple input files that share the same column names. In addition I've separated passing columns as variable names/subscript names so it tells you on error whether it couldn't find a subscript or data variavle column.

tests/integration/test_time_series.py
```diff
@@ -3,25 +3,23 @@
 import pathlib
```
Dashadower, line 3: I've disabled all the tests in this file since we dropped shift for now.

tests/integration/test_multiple_data.py
```diff
@@ -104,25 +105,33 @@ def test_multiple_dataframes_eight_schools_error_to_many_to_few_rows():
...
+    y[school]' ~ normal(theta[school], sigma[school]);
+    theta[school]' = mu + z[school] * tau;
+    z[school] ~ normal(0, 1);
     mu ~ normal(0, 5);
     tau<lower = 0.0> ~ log_normal(0, 1);
     """
 
-    with pytest.raises(Exception, match="unique"):
-        Model({"y_data": y_data_df, "sigma_data": sigma_data_df}, model_string=model_string)
+    # https://stackoverflow.com/questions/70499076/how-to-check-exception-cause-using-pytest-raises
```
Dashadower, line 118: Since we're using `raise E from e` alot to pass around exceptions, pytests can't match exception messages from the inner cause exceptions.

In addition, I've found there's too many intermediate phases in the stack trace, which sorta "hides" the original(cause) exception in a wall of text, potentially making it hard to find.

tests/integration/test_issues.py
```diff
@@ -4,34 +4,35 @@
 import pytest
 
 from rat.model import Model
+from rat import optimize
 
 test_dir = pathlib.Path(__file__).parent
-issue_data_dir = os.path.join(test_dir, "issue_data")
+issue_data_dir = test_dir / "issue_data"
```
Dashadower, line 10: Check out this weird path join [syntax](https://docs.python.org/3/library/pathlib.html#operators). I do not like it XD

Dashadower (approved):
Tests are passing! I've made some minor changes to the logic.

tests/integration/test_time_series.py
bbbales2: Yup and no intention of bringing back shift cause it makes assumptions about variables being stored as dataframes (instead of functions).

tests/integration/test_multiple_data.py
bbbales2: Makes sense. Looking at the code there, we could:
1. Throw the error in the bind function
2. Throw the error from code calling bind

Throwing the error in the bind gives us the most specific error information.

Throwing from the code calling the bind gives us access to the ast (bind doesn't know about this).

I believe we want both things. Without passing the ast through to bind somehow, what if we adopted an add_note strategy: https://peps.python.org/pep-0678/

The lingo then would be instead of re-raising unrecoverable errors that we just append information as best we can and let the original exception propagate. In this case we'd want a way to do the AST print without using CompileError.

tests/integration/test_issues.py
bbbales2: looool, agreed, until everyone is doing it and I get used to it, os.path.join ftw

bbbales2: But to be clear, this is just a few tests, whatever is fine, no need to go change this.

rat/compiler.py
bbbales2: I had a closer look at this. I still want to move completely to requiring subscripts.

I like that the index variables are all explicit and I think it makes the models more readable.

For instance:
```
y' ~ normal(mu, sigma);
```

If y and sigma are in a dataframe, you'll potentially get a different sigma for every y.

If just y is in a dataframe, you get one sigma.

I also do not like modifying the ast. The solution there is moving this info into the variable_table.

I'm gonna leave this for now and merge stuff since tests are passing at all. Any complaints if we remove this in future tho?

Dashadower: Ohhh shoot yeah ambiguous operations is a big nono.

I agree on going with the all subscripted variables rule. Right now there's just too much additional logic required even to let the user omit the simplest ones.

rat/grammar.ebnf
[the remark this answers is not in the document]
Dashadower: I'll open an issue for this since I dont want it to get lost.

tests/integration/test_time_series.py
Dashadower: Ah okay then we might as well remove the file completely soon.

tests/integration/test_multiple_data.py
Dashadower: Interesting. What I think is necessary is printing the source code(where it messed up) and the actual informal error message as one.

Notes honestly sounds like a glorified 'append text to exception message' to me, but I think it will serve our purpose here(and if it didn't exist I would've proposed to just append the error meswage to the new exception message)

tests/integration/test_issues.py
Dashadower: Agreed

rat/compiler.py
Dashadower: > If y and sigma are in a dataframe, you'll potentially get a different sigma for every y.

If just y is in a dataframe, you get one sigma.

On second thought, I actually think this should be the right behavior. If sigma is in your dataframe and youre invoking it, then it becomes data. If not then its a parameter. Remember we always see rat models within the context of the supplied dataframes.

I'm not saying the onus of making well sefined code is upon the user, but if sigma was the observed errors I would've written it as `sigma_obs`.

This is a design decision and I'm still very much fine with requiring subscripts though.

PR #99, opened 2022-10-09
Add code to evaluate custom expressions

bbbales2:


rat/codegen_backends.py
```diff
@@ -1,13 +1,18 @@
+import jax
```
bbbales2, line 1: This file is becoming a bit of a catch-all. I kinda want to break it up.

rat/fit.py
```diff
@@ -139,18 +157,84 @@ def save(self, folder, overwrite=False):
...
 
 
+@dataclass
+class GeneratedQuantitiesFunctionGenerator(RatWalker):
+    variable_table: VariableTable
+    trace_table: TraceTable
+    parameters: Dict[str, Any]
+    traced_nodes: List[ast.Variable] = field(default_factory=list)
+
+    def process_node(self, node: ast.ModelBase):
```
bbbales2, line 167: I'm not a big fan of having an entrypoint here that isn't `.walk()`. I'd rather this be walk_Expression or something.

rat/fit.py
```diff
@@ -139,18 +157,84 @@ def save(self, folder, overwrite=False):
...
+        expression = (lambda: parser.parse(expression_string, start = "expression"))()
+
+        walker = SubscriptTableWalker(self.model.variable_table)
+        walker.process_node(expression)
+        trace_table = walker.trace_table
+
+        @jax.jit
+        @jax.vmap
+        @jax.vmap
+        def run(unconstrained_draws : numpy.ndarray):
```
bbbales2: One of the vmaps is for chain and one is for draw

tests/integration/test_optimize_bernoulli.py
```diff
@@ -25,6 +25,24 @@ def test_optimize_bernoulli():
...
+    mu[group] ~ normal(-0.5, 0.3);
+    """
+
+    model = rat.Model(model_string=model_string, data=data_df)
+    fit = rat.optimize(model, init=0.1)
+    mu_df = fit.expr("mu[group]")
+
+    # This is ordered from group 1 to 5 -- assumes draws are too
+    mu_ref = [0.0797716, -0.4084060, -0.8545570, -0.9920960, -0.7676070]
+    for chain, per_chain_mu_df in mu_df.groupby("chain"):
```
bbbales2, line 42: I'm thinking about keeping all the chains even for optimization.

rat/codegen_backends.py
```diff
@@ -233,3 +293,101 @@ def walk_Variable(self, node: ast.Variable):
...
+            if isinstance(node_variable, DynamicVariableRecord):
+                self.dynamic_variable_nodes.add(node)
+
+
+@dataclass
+class SubscriptTableWalker(RatWalker):
+    variable_table: VariableTable
+    trace_table: TraceTable = field(default_factory=TraceTable)
+
+    def process_node(self, node: ast.ModelBase):
```
bbbales2, line 369: Here's another sneaky non-`.walk()` entrypoint. I'd also like to replace this.

rat/codegen_backends.py
Dashadower: Outside of some special high-order functions(`vmap, jit`), aren't most of them in `jax.numpy`?

Dashadower:
Looks like wee need to wrap up #95 since this is dependent on it.

rat/exceptions.py
```diff
@@ -49,3 +49,8 @@ def __init__(self, operation_message: str, message: str, node: ast.ModelBase = N
 class CompileError(AstException):
     def __init__(self, message: str, node: ast.ModelBase = None):
         super().__init__("compiling code", message, node)
+
+
+class ExecuteException(AstException):
```
Dashadower, line 54: Sounds like `RuntimeException` to me

rat/fit.py
```diff
@@ -139,18 +157,85 @@ def save(self, folder, overwrite=False):
...
     Stores optimization results
     """
 
-    def __init__(self, draw_dfs):
+    def __init__(self, model: Model, draw_dfs: Dict[str, pandas.DataFrame], unconstrained_draws: numpy.ndarray):
+        self.model = model
         self.draw_dfs = draw_dfs
+        self.unconstrained_draws = unconstrained_draws
+
+    def expr(self, expression_string: str):
```
Dashadower, line 204: This should be in `Fit` since we also want this functionality for sample too.

rat/fit.py
```diff
@@ -31,6 +43,21 @@ def _read_folder_into_dict(folder: str):
     return dfs
 
 
+def _build_constrained_df(constrained_variable: numpy.array, base_df: pandas.DataFrame):
```
Dashadower, line 46: Looks like we can create a post-processing class that takes a `Fit` object that would evaluate any generated quantities and then create a single dataframe.

I dunno. I like the idea of having some kind of "toolchain" that takes some object as input, does only whatever it is designed to do, and then pass it into the next class in the queue instead of one giant class that does everything inside.

rat/fit.py
Dashadower: Not optimal, but I think it's fine if you make the notion of it being an entrypoint clear.

rat/fit.py
Dashadower: ngl that's super confusing syntax D:

bbbales2: Yeah, I don't think there's much to be done about the decorators, but in the long run I'd like to make this a regular function and not be inlining it to make the inputs and outputs clear.

rat/fit.py
bbbales2: My suspicion is that I'm trying to do too many things with one walker and I should try to separate these. I'm not sure, but all the other walkers are the same. I'd like to not change that pattern.

rat/codegen_backends.py
bbbales2: Yah, but I meant more about the contents of codegen_backends. This file is becoming a dumping ground for walkers that are used in more than one place and the file is getting kinda long. The relationships between things within this file isn't clear.

rat/fit.py
bbbales2: Good point I think I agree. I think breaking the sample/optimize functions out from the model class was a good idea because it makes what happens in any given class simpler.

Even though optimize and sample are only used in one place, it was more confusing baking them into the model.

I guess this is what they call coding to interfaces.

bbbales2:
> Looks like wee need to wrap up https://github.com/bbbales2/regressions/pull/95 since this is dependent on it.

Agreed, that's kinda been open too long.

The things left before that is effectively at feature parity with main are ifelse and recursive assignment.

We can either drop those features for now and merge into main or try to add those features now. I could go with either, but I agree too much branching (like what I'm doing here) is kind of confusing.

rat/codegen_backends.py
Dashadower: Ah yes agreed

Dashadower:
> > Looks like wee need to wrap up #95 since this is dependent on it.
> 
> Agreed, that's kinda been open too long.
> 
> The things left before that is effectively at feature parity with main are ifelse and recursive assignment.
> 
> We can either drop those features for now and merge into main or try to add those features now. I could go with either, but I agree too much branching (like what I'm doing here) is kind of confusing.

I think it's worth giving those 2 features a shot. Especially ifelse is tad too important to drop out. 

My resolution is we set a soft 'deadline' until which we try and make it work, and if we can't let's reassess.

Issue #90, opened 2022-05-14
Parameter not allowed in ifelse


bbbales2:
I think this model should work:
```
hadiff' ~ normal(skill[hteam, hnumber] - skill[ateam, anumber], sigma);
skill[team, gnumber]' = (
    skill[team, shift(gnumber, 1)] +
    epsilon[team, gnumber] * ifelse(gnumber == 0, tau0, tau)
);
epsilon[team, gnumber] ~ normal(0.0, 1.0);
tau<lower=0.0> ~ exponential(1.0);
tau0<lower=0.0> ~ exponential(10.0);
sigma<lower=0.0> ~ exponential(1.0);
```

Example data:
```
date,hteam,ateam,hscore,ascore,hadiff,hnumber,anumber
2019-10-22,TOR,NOP,130.0,122.0,8.0,0,0
2019-10-22,LAC,LAL,112.0,102.0,10.0,0,0
2019-10-23,CHA,CHI,126.0,125.0,1.0,0,0
2019-10-23,IND,DET,110.0,119.0,-9.0,0,0
2019-10-23,ORL,CLE,94.0,85.0,9.0,0,0
2019-10-23,BKN,MIN,126.0,127.0,-1.0,0,0
```

Error message:
```
rat.exceptions.ParseError: An error occured while parsing the expression at (3:29):
    epsilon[team, gnumber] * ifelse(gnumber == 0, tau0, tau)
                             ^~~~~~
Parameters are not allowed in ifelse conditions
```

Dashadower:
is gnumber a parameter? Weird how it's  being used as a subscript though.

Dashadower:
Hmm this is weird. The check for parameters in ifelse are done at the parsing stage, but we can't determine gnumber is a subscript until variable table is constructed. I think the check needs to be moved to pre-codegen check and it should work.

PR #101, opened 2022-11-01
Allow expressions on the LHS of sample statements

Dashadower:
Addresses #90

### DIFF
--- a/rat/grammar.ebnf
+++ b/rat/grammar.ebnf
@@ -5,7 +5,14 @@
 
 start::Program = { statements+:statement }+ $;
 
-statement::Statement = left:factor ~ op:('~' | '=') ~ right:expression ';';
+statement =
+    | sample_statement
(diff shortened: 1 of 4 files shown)


rat/grammar.ebnf
```diff
@@ -5,7 +5,14 @@
...
 
-statement::Statement = left:factor ~ op:('~' | '=') ~ right:expression ';';
+statement =
+    | sample_statement
+    | assignment_statement
+;
+
+sample_statement::Statement = left:factor op:("~") ~ right:expression ';';
+
+assignment_statement::Statement = left:variable op:('=') ~ right:expression ';';
```
Dashadower, line 15: The problem I see is using `variable` on LHS of statement may mean constraints are also legal? How would constraints on assigned parameters work? IIRC stan doesn't do any transformations, but just as a sanity check.

PR #102, opened 2022-11-03
RIP 1: Create workflow for creating docs branch on merge

Dashadower:
Doesn't auto-merge yet.

### DIFF
--- a/.github/workflows/create_docs_branch.yml
+++ b/.github/workflows/create_docs_branch.yml
@@ -1,0 +1,35 @@
+name: Build docs and create pull request
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  if_merged:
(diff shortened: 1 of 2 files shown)


Dashadower:
https://github.com/peter-evans/enable-pull-request-automerge#example

The auto PR merge example shows exactly what we want to do, but I can't set it up since you need a repo-bounded github PAT set up as the [repo's secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets).

bbbales2:
This code looks good to me. This means after each merge to main we should expect a docs build and PR, which makes sense.

I turned off require approval to merge, so feel free to merge whatever PRs without waiting on approve to get this working.

I don't think automerge is necessary. It looks tricky to configure with a lot of annoying security things to think about. No need for that mental overhead. Thanks for looking into it.

One nit might be, if two PRs merge into master, then there are two separate docs PRs -- somehow it'd be nicer too keep only the latest. I don't think it'll matter though, so not worth stressing over.

